PR #16, opened 2026-08-28
Fall back to system fonts for CJK glyphs

### LINKED ISSUE #4: japanese characters (and presumably others) are not rendered.
Image

charliealgernon:
## Summary
- Inter has no Hangul, kana, or CJK ideographs, so Korean, Japanese, and Chinese titles rendered as empty boxes (see #4).
- Load the platform's UI fonts at startup (AppleGothic / Hiragino on macOS, Malgun / Yu Gothic / YaHei on Windows, Noto CJK or Nanum on Linux) and append them as fallbacks after Inter.
- Avoids shipping Noto CJK in the binary, which would add tens of megabytes.

## Test plan
- [ ] On macOS, open a playlist or search result with Korean, Japanese, and Chinese titles and confirm the glyphs render (not tofu).
- [ ] Confirm Latin UI text is still Inter.
- [ ] `cargo test --features demo` (includes Hangul/kana/han glyph checks when the OS has a matching font).
- [ ] On a Linux machine without CJK fonts installed, the app still starts; titles in those scripts may remain missing until a font such as `fonts-noto-cjk` is present.

Fixes #4

### DIFF
--- a/src/demo.rs
+++ b/src/demo.rs
@@ -80,6 +80,9 @@ const TRACKS: &[&str] = &[
     "Sunlight",
     "My Friend the Forest",
     "Kaleidoscope",
+    "봄여름가을겨울",
+    "夜に駆ける",
+    "起风了",
 ];
 
 const PLAYLISTS: &[&str] = &[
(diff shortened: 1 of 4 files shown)


crmne:
Thanks for taking this on. #10 landed first with the same idea done by coverage probing rather than font names (so it also covers Arabic, Hebrew, Thai, the Indic scripts, and more), and it is on main now as 3f6f4cf and 90520f1. If you find a script or a platform it misses, please open an issue with the font you expected it to pick up.

PR #21, opened 2026-08-28
Give macOS a real application menu

dappermint:
On macOS the app has no menu bar to speak of. winit installs a skeleton one with an app menu and nothing else, so there is no File, Edit, Playback, Go or Window menu, and every command is reachable only by chord or by pointer.

The missing Edit menu is the part that bites hardest: without it the standard `cut:`, `copy:`, `paste:` and `selectAll:` selectors are unbound, so Cmd+C in the search field does nothing.

## Shape

`src/mac_menu.rs` builds the menu against AppKit directly, since winit offers no way to extend the one it installs.

Playback, Go and View items post into a queue that the frame loop drains into the same `Action`s the keyboard already produces. One path per command, so the menu cannot drift away from the shortcuts. Edit items go to the responder chain selectors instead, which is what makes the text-field editing keys work.

Two deliberate choices worth review:

- The menu is built once per window, in the eframe creator, not per frame.
- The handler object is leaked on purpose. It has to outlive every menu item that targets it, and the alternative is a `static mut`. There is one per process, created once.

The waker closure requests a repaint, so a menu pick is acted on immediately rather than at whatever frame the app draws next; without it a pick on an idle window sits until something else wakes the loop.

## Cost

None in dependencies. Rebased onto "Close to the tray and answer media keys on Windows and macOS", this builds against the `objc2` 0.6 and `objc2-app-kit` 0.3 that commit already brought in, adding only the `NSMenu` and `NSMenuItem` features to the existing entry. `Cargo.lock` is untouched. Everything is behind `#[cfg(target_os = "macos")]`.

## Verification

`cargo clippy --all-targets` clean with no new warnings, 38 tests pass on aarch64-darwin. The module has been running in my fork, though against objc2 0.5; the port to 0.6 here (`define_class!`, `#[unsafe(method(...))]`, the `NSEventModifierFlags` constants) is compiled and tested but I have not re-run the menu interactively since the port. Worth a click through Playback and Edit before merging.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,7 +89,7 @@ souvlaki = { version = "0.8", default-features = false }
 # Serving AppKit while no window exists (the status item lives on the main
 # thread), and leaving the Dock while hidden.
 objc2 = "0.6"
-objc2-app-kit = { version = "0.3", default-features = false, features = ["std", "NSApplication", "NSEvent", "NSResponder", "NSRunningApplication"] }
+objc2-app-kit = { version = "0.3", default-features = false, features = ["std", "NSApplication", "NSEvent", "NSMenu", "NSMenuItem", "NSResponder", "NSRunningApplication"] }
 objc2-foundation = { version = "0.3", default-features = false, features = ["std", "NSDate", "NSObjCRuntime", "NSRunLoop", "NSString"] }
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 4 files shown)


crmne:
Landed on main as 1fd9ff0, rebased by hand because lib.rs had moved under it (the repo keeps a linear history, so GitHub could not do it). Thank you. One thing to watch once you click through it: the Playback menu's Cmd+arrow key equivalents fire ahead of the focused view, so Cmd+Left/Right in the search field now skip tracks instead of moving to the line's ends, the same trade Spotify's own menu makes.

PR #22, opened 2026-08-28
Put playback on the macOS media keys

dappermint:
macOS has no desktop media controls today. The F7-F9 keys, Control Centre, the lock screen and the AirPods pinch all reach whatever else is playing, and Linux gets MPRIS while macOS gets the no-op stub.

Two commits, and the first is worth reading on its own.

## 1. Name the media-control vocabulary for what it is

`MprisCommand`, `MprisState` and `MprisTrack` were defined twice, once in `mpris.rs` and once in `mpris_stub.rs`, and every platform that is not Linux imported MPRIS-named types to say it had none.

They move to `src/media.rs`, imported by both platform modules, so there is one definition instead of two copies to keep in step. `lib.rs` routes `media_controls` per platform, `MprisService` becomes `MediaControls`, `AppOptions::mpris` becomes `media_controls`. Names only, no behaviour change, and `single_instance.rs` keeps every MPRIS D-Bus name exactly as it is because those are the wire protocol.

## 2. Put playback on the macOS media keys

`now_playing.rs` registers with `MPNowPlayingInfoCenter` and `MPRemoteCommandCenter` behind the same `MediaControls` interface `mpris.rs` implements, so `app.rs` drives both through one path and neither platform is a special case.

Title, artist, album, duration and elapsed time are published; cover art is fetched off the main thread and handed over as an `NSImage`; play, pause, next, previous, seek and scrub come back as `MediaCommand`s.

Position is published rather than pushed on a timer. The system interpolates from the elapsed time and the playback rate, so a paused track does not creep and there is no repaint budget spent on keeping the lock screen honest.

## Cost

macOS-only: `objc2`, `objc2-foundation`, `objc2-app-kit`, `objc2-media-player`, `block2`. Behind `cfg(target_os = "macos")`, so no other platform compiles or links any of it.

## The bundling catch

macOS routes the media keys to bundled applications only. This is live from `Fastpotify.app` and does nothing from `cargo run`, which is worth knowing before testing it. I added a line to the README saying so.

## Verification

`cargo clippy --all-targets --all-features` and 36 tests clean on aarch64-darwin, including a round trip through the published Now Playing info. Exercised by hand from a signed bundle: media keys, Control Centre, the lock screen, scrubbing from the system UI, and cover art all drive playback.

I cannot test the Linux side of commit 1 here. It is a rename, and `cargo clippy` passes for the tree, but the MPRIS service itself is the thing I have not run.

## Overlap

Independent of #17, #18 and #21, though #21 also adds `objc2` dependencies; whichever lands second will want its `Cargo.toml` block merged rather than added twice.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1444,6 +1444,7 @@ dependencies = [
  "aes",
  "anyhow",
  "base64",
+ "block2 0.5.1",
  "clap",
  "cpal",
  "ctr",
@@ -1465,6 +1466,10 @@ dependencies = [
  "mdns-sd",
  "memmap2",
  "mpris-server",
+ "objc2 0.5.2",
+ "objc2-app-kit 0.2.2",
+ "objc2-foundation 0.2.2",
+ "objc2-media-player",
  "open",
  "pbkdf2",
  "percent-encoding",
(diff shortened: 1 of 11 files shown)


dappermint:
Superseded by ecfd1da, which does this through souvlaki and covers Windows as well. Closing.

One piece here is still worth having and I will open it separately: media_native.rs redefines MprisCommand, MprisState and MprisTrack, which mpris.rs also defines, and lib.rs routes the Windows and macOS module under the name `mpris`. That is a small rename and de-duplication, no behaviour change.

PR #24, opened 2026-08-28
Name the media-control vocabulary for what it is

dappermint:
Follow-up to #22, which I closed: `ecfd1da` supersedes the feature, but the duplication it left behind is still worth clearing.

`MprisCommand`, `MprisState` and `MprisTrack` are defined twice, identically, once in `mpris.rs` and once in `media_native.rs`. A change to the vocabulary has to be made in both places or the two drift apart silently, since nothing checks they agree.

The naming is also now off. `media_native.rs` talks to the System Media Transport Controls on Windows and to Now Playing on macOS, neither of which is MPRIS, and `lib.rs` routes it under the module name `mpris`.

## What changes

- The three types move to `src/media.rs`, imported by both platform modules. One definition instead of two copies.
- `lib.rs` routes `media_controls` per platform, so the module name says what it is.
- `MprisService` becomes `MediaService`, `AppOptions::mpris` becomes `media_controls`.

`MediaService` rather than the more obvious `MediaControls` because souvlaki already exports a `MediaControls` that `media_native.rs` imports; the two would collide.

`single_instance.rs` is deliberately untouched. Every MPRIS bus name there is the wire protocol and has to stay exactly as it is.

## Verification

Names only, no behaviour change. `cargo fmt --check`, `cargo clippy --all-targets --all-features` and 38 tests clean on aarch64-darwin. The two remaining clippy warnings (`tray_native.rs:298`, the unused `Update` in `media_native.rs`) are already on `main` before this change; I left them alone.

Net 97 lines removed. I built the macOS side; the Linux MPRIS service and the Windows SMTC bridge I have not run, though neither has any logic touched.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -13,8 +13,9 @@ use crate::backend::{
     ApiRequest, ApiResponse, AuthStatus, Backend, Command, Event, LocalPlayback, RemoteAction,
     Waker,
 };
+use crate::media::{MediaCommand, MediaState, MediaTrack};
+use crate::media_controls::MediaService;
 use crate::model::*;
-use crate::mpris::{MprisCommand, MprisService, MprisState, MprisTrack};
 use crate::paths::AppDirs;
 use crate::player::{EngineConfig, LoadSpec, LocalState, Playback, PlayerCommand, RepeatMode};
 use crate::settings::{SessionState, Settings, ThemeChoice};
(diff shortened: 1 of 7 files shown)


crmne:
Landed on main as d80816d, rebased by hand over the lyrics commits; the two macOS warnings you noted are gone in 713c0f0. Thanks for clearing the duplication.

PR #27, opened 2026-08-28
Add full top songs page

mikenoired:
## Summary
- add a dedicated page for the user’s top songs
- link the Home page title and add a “Show more top songs” action below the 10-song preview
- keep Home to one 20-item top-tracks request; load the ranked list separately on the Top Songs page in up to two 50-item pages
- document the complete top-songs view in Everyday Use

## Validation
- cargo fmt --check
- cargo check
- cargo clippy --all-targets -- -D warnings
- cargo test --lib

### DIFF
--- a/docs/_guide/using-fastpotify.md
+++ b/docs/_guide/using-fastpotify.md
@@ -15,6 +15,11 @@ The player bar at the bottom always shows what is playing, on this
 computer *or* on any other device. Click the title to open its album, the
 artist to open the artist, the heart to save it.
 
+## Home
+
+Home previews your most-played songs. Select **Your top songs** or **Show
+more top songs** to open the complete ranked list.
+
 ## Your Library
 
 The sidebar is your library: filter it by Playlists, Albums, Artists, or
--- a/src/api/client.rs
+++ b/src/api/client.rs
@@ -874,11 +874,17 @@ impl ApiClient {
         .await
     }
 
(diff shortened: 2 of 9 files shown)


crmne:
Thanks, the page itself is right: ranked table, linked title, "Show more" only when there is more. One change before it lands, about API budget, which is the thing users hit most.

As written, every Home load asks for 50 top tracks and then 50 more, two requests where there was one, for a preview that shows ten. Please keep Home at its single 20-track request and fetch the up-to-100 list only when the Top Songs page opens (`ensure_loaded(Page::TopSongs)` asking for `TopTracks { offset: 0 }` when fewer than 100 are held, then paging as you do now). Same for `reload`: reloading the page should refetch top tracks, not all of Home.

Small things: `cargo fmt` and `cargo clippy --all-targets -- -D warnings` clean, and a line in the Everyday Use guide would be nice. Happy to merge after that.

mikenoired:
@crmne Addressed the API-budget feedback:

- Home is back to one 20-item top-tracks request.
- The full ranking is loaded only when Top Songs opens, with up to two 50-item pages.
- Reloading Top Songs now refreshes only that ranking.
- Added the Everyday Use note and ran fmt, strict Clippy, and the test suite.

crmne:
Rebased, built, clippy and tests clean, merged. Home is back to one request and the full ranking loads only on the page. Thanks for turning it around so fast.

crmne:
Merged into main as a47b1a4 (fast-forward, so GitHub did not mark it automatically). Thanks again.

PR #34, opened 2026-08-28
Control the running app from outside its window

Seigiard:
## Summary

On macOS and Windows the running Fastpotify can now be driven from outside its window: `fastpotify next`, `play-pause`, `seek`, `volume`, `shuffle`, `repeat`, `now-playing` and the rest reach the instance already running, which is enough for a launcher, a hotkey tool, or a shell script to control playback. Linux already had this through MPRIS and `playerctl`; these platforms had nothing that could address Fastpotify in particular.

## Why not an AppleScript dictionary

That is what Raycast's Spotify extensions use, and it is the obvious macOS answer. It also means an `.sdef`, an Apple event handler, and Objective-C bridging inside an egui app, for one direction of one platform. Media keys were the other candidate: they already work through `MPRemoteCommandCenter`, but they go to whichever app owns Now Playing, so they cannot name Fastpotify, and they carry no volume, seek, or shuffle.

The single-instance socket, in contrast, already exists on exactly the two platforms that lack MPRIS, already carries one verb, and already has a path into the interface. Widening it costs a parser.

## What the protocol does

A request is one `fastpotify:<verb>` line and the answer is one line. Playback verbs are acknowledged and pushed onto the same action queue the tray and the media keys feed, so they take the app's existing local-or-remote routing for free. `nowplaying` is answered from a snapshot the interface publishes each frame, so the listener thread reads no app state and takes no lock the UI holds.

Two things stay as they were. An unknown line still gets no reply at all, so a foreign program that happens to hold port 47113 is told nothing and is never mistaken for Fastpotify. And `fastpotify:show` is byte-for-byte the request and acknowledgement a second launch already sent, so instance detection across a mixed-version pair is unchanged; a new client meeting an older instance gets a clear "not running, or predates remote control" rather than silence.

Linux is deliberately left alone. It keeps MPRIS, and the subcommands there refuse and point at `playerctl` rather than growing a second way to do the same thing.

## Verification

`cargo test --features demo`, `cargo fmt --all --check`, and `cargo clippy --all-targets --features demo -- -D warnings` all pass. Five new tests cover the verb parser, the rejection of lines that are not ours, a client/listener round trip over a real loopback socket, and the command-to-action mapping including the clamp on an out-of-range volume.

Every verb was also run by hand on macOS against a build playing real music: transport, seek, volume, shuffle and repeat all took effect and the snapshot read back the new state. Windows is untested — the code path is the same `not(target_os = "linux")` one, but I have no machine to run it on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,9 @@ as one ordinary desktop application rather than a shell plugin.
 - **One instance.** Launching it again surfaces the window that is already
   open instead of starting a rival copy, on every platform.
 - **Desktop integration.** MPRIS on Linux, so media keys, the shell, and
-  `playerctl` see Fastpotify like any other player.
+  `playerctl` see Fastpotify like any other player. On macOS and Windows,
+  `fastpotify next` and its siblings drive the running app from a terminal,
+  a launcher, or a hotkey.
 
 ## Install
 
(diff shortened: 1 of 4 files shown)


crmne:
Rebased, built, clippy and tests clean (48 now), merged into main as ee338f0; fast-forward, so GitHub did not mark it. The protocol guard, the bounded read, and the round-trip test are exactly how I would want this done. Thank you.

PR #40, opened 2026-08-28
fix: enable tokio fs feature in Cargo.toml for cross-platform builds

kveld9:
## Summary
Enables the `fs` feature flag for `tokio` in `Cargo.toml`.

## Motivation
Commit `fabf009` introduced `tokio::fs` calls in `src/backend.rs` for playlist caching. On targets like Windows and macOS where `tokio/fs` is not pulled transitively via platform-specific crates, this breaks compilation with `E0433: cannot find fs in tokio`.

## Verification
- `cargo check --features demo`
- `cargo fmt --all --check`


crmne:
Thank you, this was exactly right. The same one-line fix landed on main as 6445f62 shortly before this was opened, so closing rather than merging. The broken CI that revealed it is green again.

PR #41, opened 2026-08-28
Animate hovers across cards, rows, sliders and the sidebar

zHeuzy2:
Fixes the hover-animation part of #6, addressed per review (squashed `09d6168 + bc2efe2`).

### What this does

- **Cards** (shelves, grids): lift on hover — cover grows slightly, shadow deepens, play button fades and slides up. All driven by a single `hover_t` so elements move together.
- **Track rows**: hover background fades in/out.
- **Sidebar rows**: soft pill highlight with accent edge on the active row.
- **Sliders** (seek/volume): fill colour eases between idle and accent, handle glows while active.
- **Play button** (`theme::circle_button`): grows on hover, shrinks on press, both animated.

All use `egui::Context::animate_bool_with_time` / `animate_value_with_time` helpers, no continuous repaint when idle.

### Fixes per review

- `track_row` no longer paints the old `if hovered { rect_filled(...) }` under the animated fill — only the `hover_t`-driven fill remains.
- `paint_shadow_lift` keeps the resting shadow: early-return only on `!palette.dark`, `lift` is clamped to `0..=1` so `lift=0` renders the same `alpha 120 / blur 28` as `paint_shadow`.
- Card play button size is not inverted: removed `entrance = 1.0 - 0.2*lift` and kept a constant `44.0`.
- Sidebar row: `animate_bool(true)` on a fresh `response.id` is a no-op in egui. Now always calls `animate_bool_with_time(response.id.with(...), response.hovered(), 0.12)` and only paints when `hover_t > 0`, with `ease_out`.
- Animations keyed on `response.id` (not `ui.id()`) so sibling widgets don't share one animated value.
- Removed duplicate `Pencil` icon entry added in the original commit.

### Screenshots

Default demo (no hover):

hover off

Forced hover (`FASTPOTIFY_FORCE_HOVER=1` — all cards/rows lifted, play buttons visible):

hover on

Reproduce: `cargo run --features demo` and move the cursor over cards/rows; or `FASTPOTIFY_FORCE_HOVER=1 cargo run --features demo -- --demo-shot out.png --demo-shot-delay 9000` for a headless hover capture.

### Checks

- `cargo fmt`
- `cargo clippy --all-targets --features demo -- -D warnings` — clean
- `cargo test` — 50 passed
- No continuous `request_repaint` when idle

### Notes

Second of the split requested in https://github.com/crmne/fastpotify/pull/6#issuecomment-5451192200. Tint/page/toast fades are not included.

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -6,7 +6,7 @@
 //! through a [`Palette`] so light and dark stay coherent and album-art tints
 //! can be blended in without hunting for hard-coded colours.
 
-use egui::{Color32, CornerRadius, Response, Sense, Stroke, Vec2};
+use egui::{Color32, CornerRadius, Response, Rgba, Sense, Stroke, Vec2};
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub struct Palette {
@@ -346,7 +346,9 @@ pub enum Icon {
     Loader,
     Lock,
     LogOut,
+    Maximize2,
     Mic,
+    Minimize2,
     Minus,
     Monitor,
     Moon,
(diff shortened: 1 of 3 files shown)


crmne:
I cannot tell what this changes from the diff and stills cannot show motion. Could you add a short screen recording of before and after, per the pull request template? Happy to look again with that.

PR #45, opened 2026-08-28
Restore window size, position and queue panel from session

zHeuzy2:
Fixes the window-geometry part of #6, addressed per review.

### What this does

- `SessionState` now stores `window_size: Option<[f32;2]>`, `window_pos: Option<[f32;2]>` and `queue_open: Option<bool>` (all `#[serde(default)]` so old `session.json` files still load).
- `App` keeps `session_window_size/pos` (from the file, consumed on the next `attach`) and `last_window_size/pos` (updated each `frame_ui` from `viewport().inner_rect` / `outer_rect`).
- On `attach` the window is resized/moved with `ViewportCommand::InnerSize` / `OuterPosition`, but only if the size is within `400..3000 × 300..2000` and the position is inside `[-1000,5000]` on both axes — so a window that was on a now-disconnected monitor does not open off-screen. On Wayland `OuterPosition` is a no-op, which is fine.
- Geometry is updated every frame with the actual rects the compositor reports, not with constants duplicated from `main.rs`. No coupling to any mini-player.
- On exit `save_state()` writes `last_window_size.or(session_window_size)` etc. and `queue_open`.

### Fixes per review

- Without mini-player coupling.
- Restored position is validated against a large virtual desktop (covers multi-monitor disconnect).
- No `760×520 / 1240×800` constants.

### Checks

- `cargo fmt`
- `cargo clippy --all-targets --features demo -- -D warnings` — clean
- `cargo test` — 50 passed
- No continuous repaint; geometry only updates when the viewport reports a rect

### Notes

Fourth of the split requested in https://github.com/crmne/fastpotify/pull/6#issuecomment-5451192200. Mini-player, tint/page/toast fades, preamp and `dev.ps1` are not included. `main` currently fails on Windows (`tokio::fs` without the `fs` feature) — not addressed here to keep the diff focused.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -211,6 +211,12 @@ pub struct App {
     remote_recheck_at: Option<Instant>,
     pub seek_preview: Option<f32>,
     pub volume_preview: Option<f32>,
+    /// Window geometry to restore on next attach, from the session file.
+    session_window_size: Option<[f32; 2]>,
+    session_window_pos: Option<[f32; 2]>,
+    /// Last observed window geometry, updated each frame for saving.
+    last_window_size: Option<[f32; 2]>,
+    last_window_pos: Option<[f32; 2]>,
     last_eviction: Instant,
     pub sign_in_url: Option<String>,
     /// The Web API application the current sign-in belongs to, so Settings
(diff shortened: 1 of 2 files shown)


crmne:
Landed on main as 2453b0c. The session file had grown other fields since you branched; both sets live together now. Thank you, and the off-screen clamping was thoughtful.

PR #46, opened 2026-08-28
feat: make queue side panel resizable and optimize compact row layout

kveld9:
## Summary
- Makes the Queue side panel resizable with width persistence in `Settings`.
- Optimizes the layout of compact track rows by removing unnecessary fixed padding for hidden action buttons, more than doubling available text width for titles and artists.

## Motivation
The Queue side panel had a hardcoded fixed width of 360px. When making it resizable, compact track rows previously reserved ~148px statically for hover-only buttons, creating an awkward empty gap and severely truncating song titles.

## Changes
- Added `queue_width: f32` to `Settings` (default `360.0`).
- Configured `queue-panel` in `src/ui/queue.rs` with `.resizable(true)`, range `280.0..=560.0`, and setting persistence.
- Refined `columns` and row layout in `src/ui/widgets.rs` for `row.compact`.

## Verification
- `cargo fmt --all --check`

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -46,6 +46,7 @@ pub struct Settings {
     /// Last local volume, 0..=65535.
     pub volume: u16,
     pub sidebar_width: f32,
+    pub queue_width: f32,
     pub search_history: Vec<String>,
     pub show_shortcut_hints: bool,
     /// A personal Spotify Web API application id, if the user registered one.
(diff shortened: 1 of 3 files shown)


crmne:
Landed on main as 7784b57. Main's row code had grown an Added By column since you branched, so I resolved the overlap while landing it; the compact-row work and the resizable panel came through intact. Thank you.

PR #48, opened 2026-08-28
feat: make lyrics side panel resizable and persist width

kveld9:
## Summary
Makes the Lyrics side panel resizable (range 280px to 640px) and persists its width in `Settings`.

## Motivation
The Lyrics panel had a fixed hardcoded width of 360px. On wide screens or during songs with long lyric lines/verses, lines would wrap aggressively into multiple lines. Making it resizable allows users to adjust it to their preferred reading size.

## Changes
- Added `lyrics_width: f32` to `Settings` (default `360.0`).
- Updated `side_panel` in `src/ui/lyrics.rs` to `.resizable(true)` with range `280.0..=640.0` and width change persistence.

## Verification
- `cargo fmt --all --check`

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -46,6 +46,7 @@ pub struct Settings {
     /// Last local volume, 0..=65535.
     pub volume: u16,
     pub sidebar_width: f32,
+    pub lyrics_width: f32,
     pub search_history: Vec<String>,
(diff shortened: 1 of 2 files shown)


crmne:
Landed on main as 9b4ad25, unchanged. Thank you.

PR #49, opened 2026-08-28
Add resilient alternate local playback with Piped and yt-dlp

Master0fFate:
## Why this change

Fastpotify currently uses librespot for local playback, which means Spotify Connect only works for Premium accounts. This pull request adds a separate local playback path for Free accounts and for sessions where Spotify has not confirmed Premium status. Confirmed Premium accounts still use Spotify Connect by default, and Premium users can select the alternate player in Settings if they prefer it.

The alternate player uses Spotify only for authentication, library access, search, playlists, and track metadata. It resolves a separate audio match through a user-configured Piped-compatible endpoint or yt-dlp. The interface labels the source clearly, and Spotify credentials are never passed to either provider.

## How it works

The new `src/alternate` engine owns matching, queue state, downloading, decoding, output, and recovery. It supports AAC/M4A, MP3, and WAV, rejects video-only and unsupported Opus/WebM streams, and refuses candidates below a configurable confidence score. Live, remix, karaoke, cover, and duration mismatches are penalized so a weak match does not start by accident.

Playback is progressive. The downloader reads an initial prefix, probes the container, and starts decoding when enough data is available. HTTP range support is used for sparse buffering and seeking, including seeks into data that has not been downloaded yet. Rapid seeks cancel obsolete range work. Received ranges remain available across transient failures, requests retry with bounded backoff, and expired media URLs can be refreshed without repeating the search. Invalid or dishonest range responses fail instead of looping or corrupting playback state.

The local session keeps play, pause, previous, next, shuffle, repeat, seek, volume, and queue behavior inside Fastpotify. Alternate playback does not call Spotify's `/me/player` endpoints. Audio-device loss recreates the local output and resumes the same track and position. Buffering also freezes the visible playback clock rather than letting it drift ahead of the audio.

Account routing is deliberately conservative. Only an explicit `product=premium` result can start librespot. Free, open, missing, and unknown product values use the alternate engine, which avoids the process exit librespot can trigger for unsupported accounts. Switching playback modes stops the other engine before starting the selected one.

## yt-dlp packaging

Release builds can embed one official yt-dlp executable for the target platform. The version and SHA-256 values are pinned in `third_party/yt-dlp/manifest`. The release workflow downloads the asset from the official yt-dlp release, verifies both the local pin and upstream checksum, and then builds Fastpotify. The application never downloads yt-dlp at runtime.

At startup, the embedded executable is verified and extracted to Fastpotify's state directory. A user-installed yt-dlp is preferred only when its parsed version is strictly newer than the bundled version. The executable itself remains ignored in Git; the repository contains the pin, vendor scripts, license, and notice.

## Interface and documentation

Settings now includes playback-source selection, a Piped API base URL, an optional yt-dlp path, match confidence, and skip-on-miss behavior. The device picker and top bar distinguish alternate local playback from Spotify Connect, and long source labels cannot push the avatar outside the window. The README and reference documentation describe storage, provider traffic, formats, recovery behavior, and the limits of the alternate player.

## Verification

The full test suite passes with 206 tests, including progressive MP3/M4A playback, sparse and rapid seeking, connection loss and recovery, malformed HTTP ranges, expired URL refresh, decoder starvation, audio-device loss, account routing, queue behavior, source-label layout, and bundle extraction. I also ran:

```text
cargo test --all-targets --features demo
cargo check --release
git diff --check
```

The repository diff and history were checked for OAuth tokens, credentials, private keys, logs, settings files, and other runtime state. None are included in this branch. The vendored yt-dlp binary is ignored and is not part of the commit.

PS: I checked the current platform terms before opening this pull request. The implementation does not decrypt, download, or proxy Spotify audio, make a Free account appear Premium, or send Spotify credentials to another provider. In that technical sense, it does not bypass Spotify's audio access controls. I cannot honestly claim blanket terms approval, however. Spotify's current [Developer Policy](https://developer.spotify.com/policy) says developers must not create a product "integrated with streams or content from another service," and [YouTube's Terms](https://www.youtube.com/static?template=terms) restrict automated access and downloads unless the service or rights holder authorizes them. This change is therefore policy-sensitive despite the separation from Spotify audio. Maintainers should review that point before accepting it, and users remain responsible for the providers and content they choose.


crmne:
This is careful work and it reads like you enjoyed building it, but it is out of scope by the project's written boundary: Spotify tracks come from Spotify, and Fastpotify does not substitute audio from other catalogues. CONTRIBUTING.md has the full reasoning. Closing for scope, not for quality.

PR #50, opened 2026-08-28
Add a persistent sidebar toggle

jdubba1:
## Why

I have a personal vendetta against the library sidebar, so I've added functionality to hide it via cmd+B.

## What changed

- Add Cmd/Ctrl+B to show or hide the sidebar
- Persist sidebar visibility across restarts
- Add top-bar and native macOS View menu toggles
- Show an icon-only Home shortcut while the sidebar is hidden
- Cover settings compatibility and hidden-sidebar rendering

## Testing

- cargo fmt --check
- cargo test --features demo,tokio/fs
- cargo clippy --all-targets --features demo,tokio/fs -- -D warnings
- Manually tested shortcut, toolbar, View menu, restart persistence, and resizing on macOS

## Note

Current main uses tokio::fs without enabling Tokio's fs feature on macOS. Local validation enabled that dependency feature at invocation time; this PR does not include an unrelated Cargo change.


crmne:
Landed on main as 18dec25, essentially unchanged, including the demo coverage. Thank you.

PR #52, opened 2026-08-28
feat: add sticky collection headers

mikenoired:
image

## Summary

Adds Spotify-style sticky collection headers that keep playback controls available while scrolling.

- Adds animated sticky headers for Liked Songs, playlists, albums, and artist pages.
- Uses the page’s original background gradient mesh, keeping the sticky header visually continuous with the top bar.
- Adds compact Artist sticky headers that play the artist’s Top Songs as a queue.
- Keeps Play/Pause behavior consistent with the main player: pausing and resuming preserves the current position.
- Keeps the floating scrollbar below the visible sticky header.
- Preserves existing table sorting and header behavior for collection pages.

## Validation

- `cargo check`
- `cargo test` (51 tests)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -2441,6 +2441,8 @@ impl App {
                 shuffle: shuffle_first.then_some(true),
                 at: Instant::now(),
             });
+        } else {
+            self.assumed_context = None;
         }
         match self.target() {
             Target::Local => {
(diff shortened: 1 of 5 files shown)


crmne:
Thank you, this is cleanly done. I sat with it and decided against sticky headers: five columns are graspable without them and the header row would spend vertical space on every scroll. Closing as a product decision, not a quality one.

PR #53, opened 2026-08-28
Round the Liked Songs cover

mikenoired:
## Summary
- render the Liked Songs gradient as a cached texture
- use the native cover corner radius, matching album and playlist artwork
- remove the mesh workaround that left artifacts in rounded corners

## Testing
- `cargo check`
- `cargo test` (51 tests)


crmne:
Landed on main as 23018cf, unchanged. Thank you.

PR #54, opened 2026-08-28
Preserve the Liked Songs counter

mikenoired:
## Summary
- keep the Liked Songs count visible while the list refreshes
- preserve the counter during asynchronous track updates

## Testing
- `cargo check`
- `cargo test` (51 tests)


crmne:
Landed on main as 718c583, unchanged. Thank you.

Issue #36, opened 2026-08-28
Invalid Spotify Client ID can trap the user on the login screen


TnZzZHlp:
### Fastpotify version and how you installed it

0.2.0

### Operating system

Windows

### System details

Windows 11

### Spotify plan

Premium

### What did you do, and what happened instead?

1. Signed in to Fastpotify normally
2. Opened Settings → "Make it even faster"
3. Entered an incorrect/invalid Spotify Client ID
4. Clicked "Switch now"
5. Spotify authentication failed and Fastpotify returned to the login screen
6. The application is now stuck on the login screen, with no way to return to Settings to correct or clear the Client ID

Expected behavior:
The login screen should provide a way to change/reset the custom Client ID or switch back to the default shared app

Currently, recovery requires manually editing Fastpotify's configuration files outside the application

### fastpotify.log

```text
No relevant error was recorded in fastpotify.log. The issue is that after switching to an invalid Client ID, the UI is stuck on the login screen and Settings are no longer accessible
```

crmne:
Fixed on main (f764090): a failed sign-in now offers one button, Use the shared Spotify app instead, which clears the personal Client ID and signs in with the shared app. No more editing files by hand. Ships in 0.2.1.

PR #55, opened 2026-08-28
Use shared and personal Spotify Web API apps together

TimotheFCN:
## Why

Fastpotify normally uses a shared Web API app. Users can also configure a personal Spotify Development Mode app to get a separate API quota.

Previously, configuring a personal app meant relying on it for every request. That caused Spotify-owned playlists and some catalog data to disappear because Spotify restricts those endpoints in Development Mode.

This change uses both Web API apps together. Fastpotify keeps the shared app for complete Spotify coverage and sends supported requests through the personal app when one is configured.

## What changed

A new API gateway routes each Spotify Web API request to either the shared app or the optional personal app.

The shared Web API app handles:

- Complete playlist library and search results
- Spotify-owned and external playlists
- Recommendations, artist top tracks, and related artists
- Saved-state checks and follows that name a playlist
- Account data unavailable through Development Mode

The personal Web API app handles supported requests such as:

- Playback, devices, and queue
- Library reads and writes
- Playlist creation
- Owned or collaborative playlist changes
- Supported catalog requests

When Fastpotify does not yet know who owns a playlist, its requests go through the shared app, which can serve any playlist. Once a response reveals the owner, later requests for that playlist move to the personal app. No extra request is spent on classification.

Each app has its own token, concurrency limit, cooldown, and rate-limit state. A `429` response from one app does not pause requests through the other, and `Retry-After` waits are capped at 30 seconds.

## Resilience

- A transient network failure during token refresh keeps the current session; only Spotify rejecting the grant signs a session out.
- Launching offline shows "Connecting" and retries verification with backoff until the network returns — no forced re-authorization.
- A fresh browser grant is saved to disk immediately, so a failed or rate-limited verification never discards it.
- The playlist disk cache is only shown after Spotify's snapshot confirms it is current; a stale cache is discarded, never shown.
- If the two grants belong to different Spotify accounts, the sign-in fails with a visible error instead of hanging.

## User-visible behavior

Adding a personal Spotify app now improves API capacity without removing content available through the shared app.

Spotify-owned playlists remain visible and can be opened. Playlist search remains complete, and artist discographies load within Spotify's current endpoint limits.

Late responses from older page requests can no longer replace newer content. Playlist caches are also scoped to the signed-in Spotify account.

## Upgrades

Existing Web API tokens migrate automatically:

- Tokens created with Fastpotify's shared app become the shared Web API session.
- Tokens created with a custom Client ID become the personal Web API session.
- Users upgrading from a custom-app-only setup keep their personal authorization and are prompted once to authorize the shared app.
- Signing out removes both Web API grants.

The separate local playback credential is unchanged.

## Code guide

- `src/api/gateway.rs` contains the routing rules, session states, and the playlist-ownership map.
- `src/auth.rs` handles authorization and migration of existing tokens.
- `src/backend.rs` manages both sessions and their lifecycle.
- `src/api/client.rs` contains per-session limits, cooldowns, and Spotify endpoint handling.
- `src/app.rs` handles request generations and account-scoped playlist caching.

## Verification

- `cargo test --features demo`
- `cargo fmt --all --check`
- Production documentation build
- Fresh shared and personal app authorization
- Spotify-owned playlist discovery, opening, and playback
- Artist discography loading

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ librespot-protocol = { version = "0.8", default-features = false }
 rodio = { version = "0.21", default-features = false, features = ["playback"] }
 cpal = "0.16"
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-native-roots", "json", "gzip", "http2", "blocking"] }
-tokio = { version = "1", features = ["rt-multi-thread", "sync", "time", "macros"] }
+tokio = { version = "1", features = ["rt-multi-thread", "sync", "time", "macros", "fs"] }
 sha1 = "0.10"
 sha2 = "0.10"
 base64 = "0.22"
--- a/README.md
+++ b/README.md
@@ -128,10 +128,11 @@ here, set up once") or Settings; it needs Spotify Premium, and librespot
 stores a reusable credential so it also never asks again. Browsing and
 remote control work on any account without this step.
 
-By default the Web API uses the shared public application also used by
-spotify-player, ncspot, and Omarchy Spotify. If you hit rate limits you can
-register your own (free) Spotify application and paste its Client ID in
-Settings → Account.
+The Web API always keeps shared catalog coverage. You can also register a
+personal Spotify Development Mode app and paste its Client ID in Settings →
+Account; supported requests use its separate quota while complete playlist
+views, playlist-bearing search, external playlists, and unavailable operations
+continue through the shared app.
 
 ## Keyboard shortcuts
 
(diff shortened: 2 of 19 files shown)


TimotheFCN:
This also fixes https://github.com/crmne/fastpotify/issues/36 indirectly

crmne:
I really like this: keeping the shared app for coverage and spending the personal quota where Development Mode allows is the right architecture, and it attacks the rate limiting people actually hit. It is a big change, so it gets a careful review. Fair warning that main moved a lot this week (playlist caching, request plumbing), so a rebase will likely be needed; I will review on this branch first and flag anything structural.

crmne:
Landed on main as e5414cd and d930b77. This is the best-engineered contribution the project has received: the routing gateway, the same-account verification, the per-account cache directories, the generation tags, and the quota-versus-burst distinction all survived review untouched. I resolved it against a week of drift on main (your artist-albums profile cap replaced an interim retry of mine, and the sidebar toggle and login escape kept their places) and the docs took your wording. Thank you.

PR #58, opened 2026-08-29
Widen the control channel to what a button needs

oyuh:
Hey! Ran into this while wiring Fastpotify up to a Stream Deck, but it turned
out not to be Stream Deck specific at all. It's the same control channel the
Raycast/Alfred scripts already use.

The problem: the existing verbs all came from media keys, so they only toggle.
There's no way to say "shuffle on", only "shuffle, change". So anything drawing
the current state on a button drifts out of sync with Spotify the second it
misses an update, and the only way to fix it is to press it again.

This adds the missing half:

`shuffle-set on|off`, `repeat-set off|context|track`, `seek-to <ms>`,
`save-toggle`, `play-uri <uri>`, `transfer <device-id>`, and a `devices` read.
They all map to `Action`s that already existed, so nothing new actually happens
to playback. They're subcommands too, so `fastpotify like` and friends work
from a launcher or a hotkey.

The now-playing snapshot also picks up three fields: artwork URL, whether the
track is saved, and which device the sound is coming out of. They're appended
after the original nine, so anything reading the old shape is unaffected.
`saved` is `yes`/`no`/`unknown`, and it's free, since the playing track already
gets looked up through `request_contains`. No extra request.

`devices` answers as JSON rather than tab-separated, since people name their
speakers whatever they like and a separator won't survive that. One quirk worth
knowing about: the app only refreshes that list while its own picker is open,
so the read kicks off a refresh as well. That means a cold first call comes
back empty and the next one has it. There's a comment at the site explaining
that.

`play-uri` and `transfer` are the only verbs carrying free text, and anything
on the machine can reach the port, so both check their argument before the app
sees it.

**Heads up: the first commit is unrelated.** `fabf009` added `tokio::fs`
without declaring the feature that compiles it. Linux gets it for free from
`mpris-server`, so the build there never noticed, but Windows won't compile at
that commit and macOS doesn't have that dependency either, so I think CI is red
on both right now. It's one word in `Cargo.toml`. I put it first so every
commit on the branch builds on its own, but happy to split it into its own PR
if you'd rather take it separately.

fmt, clippy (`-D warnings`) and the 54 tests all pass, and both commits build
standalone. Tested against a real account too. The snapshot reads back
correctly and `devices` does the empty-then-populated thing described above.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,10 @@ librespot-protocol = { version = "0.8", default-features = false }
 rodio = { version = "0.21", default-features = false, features = ["playback"] }
 cpal = "0.16"
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-native-roots", "json", "gzip", "http2", "blocking"] }
-tokio = { version = "1", features = ["rt-multi-thread", "sync", "time", "macros"] }
+# `fs` is used by the playlist cache in `backend.rs`. On Linux the
+# mpris-server dependency happens to turn it on, which is why leaving it out
+# built there and nowhere else.
+tokio = { version = "1", features = ["rt-multi-thread", "sync", "time", "macros", "fs"] }
 sha1 = "0.10"
 sha2 = "0.10"
 base64 = "0.22"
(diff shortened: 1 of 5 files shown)


crmne:
Landed on main as 2ad8d18 and 158fca2. This is exactly how a contribution to the control channel should look: state verbs over toggles, appended raw fields so old scripts keep reading, and you even diagnosed the tokio fs unification independently. I adjusted the README wording slightly while resolving against main. Thank you.

PR #59, opened 2026-08-29
macOS: hide the titlebar and run the UI to the top edge

koels:
## macOS: hide the titlebar and run the UI to the top edge

On macOS the app currently sits under the stock titlebar — a separate band of chrome with the window title in it, above the sidebar and the top bar. It reads as a generic window rather than as a music player: Spotify, Music, and most other players on the platform hide that strip and let their own chrome run to the top edge with the traffic lights floating over it.

This does the same.

### What changed

- **`main.rs`** — the viewport turns on `fullsize_content_view` and turns off `titlebar_shown` / `title_shown`. macOS-only options; other platforms are untouched.
- **`theme.rs`** — new `titlebar_inset(ctx)`, the vertical space the traffic lights need. It is `28.0` on macOS and `0.0` everywhere else, and also `0.0` in fullscreen, where the buttons are gone and the space would just be padding.
- **`ui/sidebar.rs`** — the panel's top margin picks up the inset, so the first nav row starts below the traffic lights instead of behind them.
- **`ui/topbar.rs`** — the bar grows upwards by the inset into the space the titlebar used to occupy. The row's own layout is unchanged.
- **`ui/mod.rs`** — new `titlebar_drag`, which gives back what the titlebar was doing: dragging the empty area moves the window. It is registered *before* the widgets that sit on top of it, so the nav arrows, the search field, and the account button keep their own clicks. It is a no-op wherever the inset is zero.
- **`ui/login.rs`** — the same drag region on the sign-in background, so the window is still movable before you are signed in.

### On the drag

The command goes out on the **press**, not on egui's drag threshold. AppKit begins the move from the mouse-down event that is still live; by the time the threshold is crossed that event is stale, the window stays put, and AppKit logs:

```
Warning: Window move completed without beginning
```

That is what the first cut of this branch did, and it is fixed here.

The consequence is that there is **no double-click-to-zoom** on the bar. Taking the press hands the rest of the gesture to AppKit's native drag loop, so a second click never comes back to egui. I tried it both ways; a reliable drag seemed clearly worth more than the zoom shortcut, but say the word if you'd rather have the trade the other way.

### Testing

Built and run on macOS (`--demo`), with synthesized input against the real window:

- Dragging the empty part of the top bar moves the window by exactly the drag delta, with no AppKit warning logged.
- A press-and-move that *starts on the search field* leaves the window where it is — widget priority holds, so the arrows, search, and account button are unaffected.
- The sidebar and top bar render correctly against the top edge with room left for the traffic lights.

### One thing to flag

`main` does not currently build for me, on a line unrelated to this PR — `src/backend.rs:1152` calls `tokio::fs`, but the `tokio` dependency in `Cargo.toml` does not enable the `fs` feature:

```
error[E0433]: cannot find `fs` in `tokio`
    --> src/backend.rs:1152:35
```

I confirmed that on a clean checkout before making any changes. I built and tested this branch with `tokio`'s `fs` feature enabled locally, then reverted that, so this PR stays scoped to the titlebar. Happy to fold the one-word `Cargo.toml` fix in here if you'd rather have it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -546,6 +546,13 @@ fn native_options(fullscreen: bool, mini: Option<MiniWindow>) -> eframe::NativeO
             }
         }
         None => viewport
+            // macOS: no title bar strip above the app. The content runs to
+            // the top edge and the traffic lights float over it, the way
+            // every other music player on the platform looks; the interface
+            // leaves room for them with `theme::titlebar_inset`.
+            .with_fullsize_content_view(true)
+            .with_titlebar_shown(false)
+            .with_title_shown(false)
             .with_inner_size([1240.0, 800.0])
             .with_min_inner_size([760.0, 520.0])
             .with_fullscreen(fullscreen),
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -102,6 +102,19 @@ pub const COMPACT_ROW_HEIGHT: f32 = 48.0;
 pub const PLAYER_BAR_HEIGHT: f32 = 88.0;
 pub const TOP_BAR_HEIGHT: f32 = 56.0;
 
+/// macOS hides the titlebar and draws the window content all the way to the
+/// top edge, so whatever sits at the top of the window has to leave room for
+/// the traffic lights. Zero everywhere else, and in fullscreen, where the
+/// buttons are gone.
+pub fn titlebar_inset(ctx: &egui::Context) -> f32 {
+    if cfg!(target_os = "macos") && !ctx.input(|input| input.viewport().fullscreen.unwrap_or(false))
+    {
+        28.0
+    } else {
+        0.0
+    }
+}
+
 const INTER_MEDIUM: &str = "inter-medium";
 const INTER_SEMIBOLD: &str = "inter-semibold";
 const INTER_BOLD: &str = "inter-bold";
--- a/src/ui/login.rs
+++ b/src/ui/login.rs
@@ -14,6 +14,7 @@ pub fn show(app: &mut App, ui: &mut egui::Ui, connecting: bool) {
         .frame(Frame::new().fill(palette.window))
         .show(ui, |ui| {
             let rect = ui.max_rect();
+            super::titlebar_drag(ui, rect);
             let top = super::blend(palette.window, palette.accent, 0.10);
             super::widgets::paint_vertical_gradient(ui, rect, top, palette.window);
             let card_width = 440.0;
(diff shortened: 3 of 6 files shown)


crmne:
This is one I want, but titlebar work lives or dies on how it looks. Could you add before and after screenshots, per the pull request template, including one with a playlist scrolled under the top edge? With those I will test it on a Mac and merge.

koels:
Screenshots as asked, and the `tokio` fix folded in.

**On the pull request template:** I could not find one — there is no `.github/PULL_REQUEST_TEMPLATE.md`, nor an equivalent in `docs/`, at the root, or in a `crmne/.github` repo, as of `7d9d29d`. I followed the screenshot convention the repo documents in `docs/_reference/settings-and-files.md` instead: demo mode, `--demo-page playlist:pl1`. Point me at the template and I will redo these to match it.

## Screenshots

All four are the same window at the same size (1800×1063) on the same demo playlist, captured as real windows so the titlebar and traffic lights are genuine — `--demo-shot` renders the frame buffer only and so cannot show window chrome. The two scrolled shots are at an identical scroll position.

### Playlist, at the top

**Before** — the titlebar band, with the window title, above everything:

Before: a titlebar strip labelled Fastpotify sits above the sidebar and top bar

**After** — the sidebar and top bar run to the top edge, traffic lights over the sidebar:

After: no titlebar strip; the traffic lights float over the top-left of the sidebar

### Playlist scrolled under the top edge

This is the one that matters for the inset. Same scroll position in both.

**Before:**

Before: the track list scrolled, with the titlebar band still above the top bar

**After:**

After: the track list scrolled, the top bar sitting flush against the top edge

The track rows pass under the top bar and are clipped by it exactly as they were before — the scroll area is still below the bar, so the only thing that moved is where that bar sits. Nothing scrolls behind the traffic lights, and the first sidebar row clears them.

## The `tokio` fix

Folded in as `ed81165`, stacked on top of the titlebar commit rather than squashed into it, since it is a separate concern and you have already read the first one. Happy to squash if you would rather land this as one.

```diff
-tokio = { version = "1", features = ["rt-multi-thread", "sync", "time", "macros"] }
+tokio = { version = "1", features = ["rt-multi-thread", "sync", "time", "macros", "fs"] }
```

`git log -S "tokio::fs" -- src/backend.rs` points at fabf009 ("Remember whole playlists on disk, keyed by snapshot") as where those calls came in; the feature gating them just never got added alongside. `Cargo.lock` is unchanged — the feature pulls in no new dependency.

This supersedes the "One thing to flag" section in the description above: `cargo build` now completes on a clean checkout of this branch with no local workaround, which also means these screenshots reproduce straight from it:

```
cargo run --features demo -- --demo --demo-page playlist:pl1
```

The titlebar commit itself is unchanged from the one you looked at.

PR #60, opened 2026-08-29
Enable tokio's fs feature for the playlist cache

kreatzzz:
Main does not compile on macOS or Windows any more. Since fabf009
("Remember whole playlists on disk, keyed by snapshot"), the playlist
cache reads and writes with `tokio::fs`, but tokio's `fs` feature is
never enabled in Cargo.toml. On Linux the symbol resolves anyway,
through a dependency that turns the feature on for the graph, so the
Linux CI job stays green while the other two fail:

```
error[E0433]: cannot find `fs` in `tokio`
   --> src/backend.rs:1171:35
```

This turns the feature on. One line; `cargo test --features demo`
passes again on all three platforms.

Verified on macOS: `main` fails `cargo check` with 7 errors, all of
them this; with the line, `cargo test --features demo` is green.


crmne:
Thank you, this was exactly right. The same one-line fix landed on main as 6445f62 before this was opened, so closing rather than merging.

PR #61, opened 2026-08-29
Translate and romanize lyric lines

kreatzzz:
Two toggles in the lyrics panel's header, next to Follow:

- **Translate** echoes every line in the reader's language, small and
  quiet under the original, so a song in words you do not speak still
  makes sense. The language is picked in Settings, forty of them, and
  English by default.
- **Romanize** rewrites the lines in Latin letters, for singing along
  to a song written in a script one cannot read: Japanese becomes
  romaji, Korean becomes romanized hangul, Cyrillic comes out Latin.

Both are off by default. Turning either on sends the song's words to
Google Translate, which answers without an account or a key; the
connections page says so plainly now, and the failure mode is a quiet
"Try again" next to the lyrics, never a broken panel.

The lyrics panel with the translation echoing each line

## How it works

One module, `src/translate.rs`, mirrors `src/lyrics.rs`: same shape,
same cache discipline, no new dependencies. The endpoint is the one
Google's own translate page uses (`clients5.google.com`, `dt=t&dt=rm`),
which answers translation and romanization together and detects the
source language for free. Three things the endpoint does made the
design:

- Translation batches well: lines joined with newlines come back with
  the newlines in place, so a song costs one or two requests, and the
  answers realign afterwards.
- Romanization loses the newlines, so it is asked for one line at a
  time, five in flight, and a chorus's repeated lines are asked for
  once.
- When the song is already in the reader's language, there is nothing
  to add, and nothing further is asked.

Answers are cached on disk for a month, next to the lyrics cache, so a
song heard again costs no request. A refused request is retried with
backoff, honouring `Retry-After`, with a dash of jitter so lines asking
together do not knock together.

## What is in the box

- `src/translate.rs` — fetch, batching, romanization, cache, and tests
  for the parsing, the alignment, the chunking, and the cache key.
- `src/app.rs` — the panel's translation follows the lyrics: asked for
  when they arrive, when the panel opens, when a toggle turns on, or
  when the language in Settings changes, and only answers for the
  current track and language are ever shown. A regression test keeps
  the next track asking for itself without a retoggle.
- `src/ui/lyrics.rs` — the two pills, a spinner in each while the
  answer is on its way, the echo line under every lyric line, and the
  romanized line standing in for the original. Settings gains a Lyrics
  section with the two switches and the language picker.
- `docs/` — the guide's Lyrics section, the connections page's
  disclosure, the settings reference, and a demo surface
  (`--demo-show lyrics`) that shows the panel with a translation on.
- This branch also carries the one-line `tokio fs` fix from #60, so CI
  is green on all three platforms; the two changes are independent and
  either can merge first.

## What I considered and did not do

- The official Cloud Translation API: clean, but needs a GCP project
  with billing attached, and romanization (v3 `romanizeText`) does not
  accept API keys at all — it would force a service-account flow into a
  desktop app for half the feature.
- LibreTranslate and friends as fallbacks: no romanization, lower
  quality; the cache and retries cover the realistic failure modes.

`cargo test --features demo` is green, including the headless render
with both toggles on.


crmne:
This is a lot of good work, but translation and romanization pull in concerns the core client should not carry, and I want to keep the lyrics panel simple. If Fastpotify ever grows a plugin interface, this is the first thing I would want written for it. Closing for scope.

PR #70, opened 2026-08-29
Size cards to their subtitle text so hover covers it

hewel:
## Problem

On cards with a two-line subtitle (e.g. the "Made for you" playlists), the hover background ends before the text: `widgets::card` allocated a fixed `image_size + 70` height, leaving 58 px under the cover for title + gaps + subtitle, while a two-line subtitle needs ≈63 px. The second subtitle line painted across the card's bottom edge — invisible until hover draws the background rect.

Before:

Screenshot from 2026-08-29 21-18-51

After:

fastpotify-pr70-after

## Fix

`card()` now lays out the title and subtitle galleys first and derives the card height from them:

- two-line subtitles stay fully inside the hover background;
- one-line subtitle cards (albums, recently played, artist grids) no longer reserve an empty second line — they get slightly shorter and the hover background hugs the text.

Shelf and grid rows keep a uniform height because their content is homogeneous (all two-line or all one-line). A mixed row (e.g. search results mixing long and short descriptions) can now size each card to its own subtitle.

No API changes; all `widgets::card` call sites are unchanged.

## Verification

- `cargo run --features demo -- --demo`, hover a "Made for you" card: both subtitle lines sit inside the hover background (screenshots above).
- One-line cards (Recently played, artist discography): hover background ends right below the subtitle instead of leaving an empty band.

## Checks

- `cargo fmt --all --check`

The jekyll docs build was not run locally (no bundler here); no docs files changed.

### DIFF
--- a/src/ui/widgets.rs
+++ b/src/ui/widgets.rs
@@ -211,10 +211,7 @@ pub fn menu_item_enabled(
         };
         let galley = ui.painter().layout_job(job);
         let pos = match halign {
-            Align::RIGHT => pos2(
-                rect.right() - 10.0,
-                rect.center().y - galley.size().y / 2.0,
-            ),
+            Align::RIGHT => pos2(rect.right() - 10.0, rect.center().y - galley.size().y / 2.0),
             _ => pos2(x, rect.center().y - galley.size().y / 2.0),
         };
         ui.painter().galley(pos, galley, color);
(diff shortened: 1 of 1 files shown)


crmne:
Merged, with one change on top of your fix: the height is now constant (title row plus two subtitle rows, taken from the fonts' row heights) rather than measured per card. Shelves lay cards out with ui.horizontal, which centres items of different heights, and Made for you mixes playlists that have a description with ones that only say "By ...", so per-card heights put the covers on different lines. Thanks for finding it, the arithmetic was clearly off.

PR #76, opened 2026-08-29
refactor(theme): separate palette colors into PaletteColors struct

### LINKED ISSUE #75: Separate the palette's colours from its structure so forks stop conflicting on theme.rs
### Before requesting

I maintain a fork of Fastpotify and merge from upstream regularly. The only
thing my fork really changes is the palette, and that one change is what
makes every merge painful: `Palette::dark()` and `Palette::light()` hold the
colours as literals, so my fork rewrites those two functions, and every
upstream commit that touches `theme.rs` lands as a conflict in exactly the
lines I rewrote. The merge for 0.3.0 conflicted there again.

Nothing else about the fork fights upstream. The palette does, because the
colours and the structure live in the same function.

This is a maintainer problem rather than an end-user feature request. I am
not asking for a theme picker in the interface, and I am not asking you to
adopt my colours.

Separate *what a colour is for* from *which colour it is*. `Palette` already
does the first job well: the field names (`surface_hover`, `on_accent`,
`dim`) are about roles, not hues, so the type is already the seam. What is
missing is a way to fill one without editing the function that defines it.

The smallest version, which is what I would propose:

- `Palette::dark()` and `Palette::light()` keep today's colours exactly, and
  stay the default. No visual change, no new setting.
- Anything that wants different colours fills a `Palette` from somewhere
  else and hands it to `theme::apply`, which already takes `&Palette`.

That alone means a fork overrides colours without touching upstream's
literals, and a merge that adds a palette field is a clean addition instead
of a conflict.

I have this running in my fork. The colours load from a file
(`<config>/themes/<name>.toml`) shaped like the flat half of an Alacritty
colour config, one `key = "#rrggbb"` per line, absent keys keeping their
built-in value, an unparseable file falling back to the default with a line
number in the log, and the file name validated before it is joined to a path
since it arrives from a hand-edited settings file. Roughly 500 lines
including tests, no new dependencies.

I am deliberately *not* proposing the file loader here. It is the part most
likely to be out of scope, and the seam is useful without it. If the split
alone is interesting I would send that as a small PR; if the loader is also
interesting I can send it separately, behind a feature flag or not at all.

Alternatives I considered:

- Keeping the fork's patch and eating the conflicts. This is what I do now.
  It is survivable but the conflict recurs on every release.
- A `--theme` flag or an env var. Same conflict, plus a new surface.
- Nothing at all: reasonable, and if that is the answer I will keep the
  patch in my fork. I would rather ask than send an unsolicited PR.

It fits as a refactor, not as a feature:

- **Focused.** The default build looks and behaves exactly as it does today.
  It adds no settings, no interface, and no modes. That is the reason I am
  pitching the split on its own rather than the theme picker.
- **Small dependency surface.** No new crates. The split is a move of
  literals; the file loader in my fork parses by hand for this reason.
- **Native and fast.** Nothing is read while drawing.
- **Cross-platform.** Colours are already platform-independent.

The honest framing: this makes life easier for people who maintain forks and
patches, and costs upstream a slightly different shape in one file. If that
trade is not one you want, closing this is a perfectly good answer and no
work is wasted.

For context on where this comes from: my fork is
[dappermint/snoop](https://github.com/dappermint/snoop), a mac-first fork
with the menu bar, media keys, and a different palette. Thank you for
Fastpotify, and for how carefully you review — the closed PRs read as more
considered than most merged ones.

kveld9:
### Summary

Closes #75.

Decouples semantic UI roles in `Palette` from raw color definitions by introducing a `PaletteColors` struct with `PaletteColors::DARK` and `PaletteColors::LIGHT` constants.

### Architectural Changes

1. **`PaletteColors` Struct**:
   - Holds the 16 raw `Color32` values.
   - Exposes `PaletteColors::DARK` and `PaletteColors::LIGHT` as `const` definitions.

2. **`Palette::from_colors` Constructor**:
   - `Palette::from_colors(dark: bool, colors: PaletteColors) -> Self` maps the color struct to the active `Palette`.
   - `Palette::dark()` and `Palette::light()` delegate to `Palette::from_colors` with their default constants.
   - `Palette::colors(&self) -> PaletteColors` extracts the active colors for round-trip ergonomics.

3. **Zero Overhead & Full Backward Compatibility**:
   - No external dependencies added.
   - Existing callers of `Palette::dark()`, `Palette::light()`, and field accessors (`palette.surface`, `palette.accent`) remain completely unchanged.
   - Forks can now instantiate custom palettes or override colors without modifying upstream's default literal definitions in `theme.rs`.

### Validation & Testing

- Added unit tests in `src/theme.rs` verifying round-trip equality for `PaletteColors::DARK`/`LIGHT` and custom color instantiation.
- Full check suite passed:
  ```powershell
  cargo fmt --all --check
  cargo test --release --features demo

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -9,6 +9,66 @@
 use egui::{Color32, CornerRadius, Response, Sense, Stroke, Vec2};
 
 #[derive(Clone, Copy, Debug, PartialEq)]
+pub struct PaletteColors {
+    pub window: Color32,
+    pub panel: Color32,
+    pub surface: Color32,
+    pub surface_hover: Color32,
+    pub surface_active: Color32,
+    pub outline: Color32,
+    pub text: Color32,
+    pub secondary: Color32,
+    pub dim: Color32,
+    pub accent: Color32,
+    pub accent_hover: Color32,
+    pub on_accent: Color32,
+    pub danger: Color32,
+    pub warning: Color32,
+    pub overlay: Color32,
+    pub shadow: Color32,
+}
+
+impl PaletteColors {
+    pub const DARK: Self = Self {
+        window: Color32::from_rgb(0x0f, 0x11, 0x14),
+        panel: Color32::from_rgb(0x15, 0x18, 0x1c),
+        surface: Color32::from_rgb(0x1d, 0x21, 0x27),
+        surface_hover: Color32::from_rgb(0x26, 0x2b, 0x33),
+        surface_active: Color32::from_rgb(0x2f, 0x35, 0x3f),
+        outline: Color32::from_rgb(0x2a, 0x30, 0x38),
+        text: Color32::from_rgb(0xf2, 0xf4, 0xf6),
+        secondary: Color32::from_rgb(0xa9, 0xb1, 0xbc),
+        dim: Color32::from_rgb(0x6e, 0x77, 0x84),
+        accent: Color32::from_rgb(0x1e, 0xd7, 0x60),
+        accent_hover: Color32::from_rgb(0x3c, 0xe8, 0x7a),
+        on_accent: Color32::from_rgb(0x0a, 0x14, 0x0e),
+        danger: Color32::from_rgb(0xf5, 0x71, 0x7f),
+        warning: Color32::from_rgb(0xf2, 0xb8, 0x5c),
+        overlay: Color32::from_rgb(0x22, 0x27, 0x2e),
+        shadow: Color32::from_black_alpha(140),
+    };
+
+    pub const LIGHT: Self = Self {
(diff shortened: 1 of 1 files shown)


vidhanio:
can we make the theme configurable via `settings.json` without having to edit source code?

crmne:
Thank you for the careful work, and I see the merge pain in #75. I am not going to carry a refactor whose reason lives in a fork, though: theme.rs is the app's, and every structure added for someone else's tree is one more thing to keep true here. A fork that only changes colours can keep its palette in one small patch on top of Palette::dark and Palette::light and reapply it; that is what those two functions are for. Closing this and #75 with that answer.

PR #114, opened 2026-08-31
Add custom colour palettes with hex/alpha support

oldirtty:
## Why

Fastpotify ships with fixed dark and light palettes. Ricers and users of tools like Noctalia/Niri who colour-match their whole desktop have no way to make Fastpotify follow a custom colour scheme, and no way to get a translucent window to fit compositor-based Wayland setups. This adds a JSON-defined palette format so the interface's colours (dark and light variants) can be customised per field, with transparency support, for blurred setups.

## What changed

- `src/palette/`: hex colour parser (`#RGB`, `#RGBA`, `#RRGGBB`, `#RRGGBBAA`), the theme file format (`ThemeFile`/`ThemeColors`, every field optional), two built-in palettes (Catppuccin Mocha, Nord), and `PaletteManager`, which resolves a theme file over the built-in dark/light palette one field at a time and supports hot reload.
- `Settings` gains `palette_file: Option<PathBuf>`, persisted and loaded like any other preference; `~` in a path is expanded against `$HOME`.
- Settings → Appearance gains a "Colour palette" section: a text field to load a theme from a JSON file, Reload (re-reads the active file from disk), Reset (back to the built-in default), and buttons for the built-in palettes.
- The remote control channel gains `palette-set <path>`, `palette-reload`, and `palette-reset` verbs, following the channel's existing scope (macOS/Windows only; Linux already routes control through MPRIS instead and is unaffected).
- The main window now allows OS compositing transparency (`with_transparent(true)`), and its clear colour follows the active palette's `window` colour, alpha included, instead of a hardcoded value. Built-in and named palettes never set alpha below 255 on `window`, so default behaviour is visually unchanged; only a custom palette that sets a translucent `window` shows the desktop behind it.

## Verification

Tested manually on CachyOS Linux (Niri WM, egui/eframe) with a custom Nord/Catppuccin-derived theme file:

- Loading a theme by path, including with a leading `~`.
- Reload after editing the file on disk, without restarting.
- Reset back to the built-in palette.
- Switching between the two built-in palettes.
- 8-digit hex alpha on `overlay`, `shadow`, and `window`, confirming the main window shows the desktop behind it when `window` is translucent, and stays opaque with the built-in palettes.
- `cargo test --locked --all-targets`: 191 passed, including new `palette::colors`, `palette::manager`, and `settings` tests.

Not exercised: the remote-control `palette-*` verbs, since those only run
on macOS/Windows.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -152,6 +152,7 @@ pub struct App {
     pub offline: bool,
     pub palette: Palette,
     applied_dark: Option<bool>,
+    pub palette_manager: crate::palette::PaletteManager,
 
     pub auth: AuthStatus,
     pub user: Option<User>,
@@ -387,6 +388,15 @@ impl App {
             .filter(|page| !matches!(page, Page::Settings | Page::Queue))
             .unwrap_or(Page::Home);
 
+        // Grabbed before `settings` moves into the struct below.
+        let palette_file = settings.palette_file.clone();
+        let palette_manager = crate::palette::PaletteManager::new();
+        if let Some(path) = &palette_file
+            && let Err(error) = palette_manager.load_from_file(path)
+        {
+            log::warn!("could not load palette from {}: {error}", path.display());
+        }
+
         let mut app = Self {
             dirs,
             settings,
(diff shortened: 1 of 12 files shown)


crmne:
Closing this in favor of #171 so theme work has one design.

PR #116, opened 2026-09-01
Fix Nix package build

mixedCase:
## Why

`nix build .#fastpotify` fails when Cargo tries to vendor the git dependencies in `Cargo.lock`. On NixOS, the MilkDrop dependency also needs X11 headers at build time and installs its library under `lib64` while `projectm-sys` expects `lib`.

## What changed

- Use `fetchCargoVendor` so the git dependencies are included in the fixed-output dependency tree.
- Add `libx11` to the Linux build inputs.
- Wrap CMake to set `CMAKE_INSTALL_LIBDIR=lib` during configuration while leaving `cmake --build` unchanged.

## Verification

- `nix flake check --no-build`
- `nix build .#fastpotify --no-link -L` (x86_64-linux; 211 tests passed)
- `nix fmt`
- `git diff --check`

- [ ] I added or updated tests for behaviour that can regress. (Packaging-only change.)
- [ ] I ran formatting, Clippy, and the relevant tests locally. (The relevant Nix package tests were run; Clippy was not run because this is a packaging-only change.)
- [x] I updated documentation for user-visible behaviour, settings, files, or network access. (Not applicable.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,6 +127,20 @@ jobs:
               Select-Object -ExpandProperty FullName
           }
 
+  nix:
+    name: Nix package
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v7
+      - uses: wimpysworld/nothing-but-nix@v9
+      - uses: cachix/install-nix-action@v31
+      - uses: nix-community/cache-nix-action@v7
+        with:
+          primary-key: nix-${{ runner.os }}-${{ hashFiles('**/flake.lock') }}
+          restore-prefixes-first-match: nix-${{ runner.os }}-
+      - run: nix build .#default
+
   docs:
     name: docs
     runs-on: ubuntu-latest
--- a/flake.nix
+++ b/flake.nix
@@ -92,6 +92,13 @@
               cargo = toolchain;
               rustc = toolchain;
             };
+            cmakeWithLibdir = pkgs.writeShellScript "cmake-fastpotify" ''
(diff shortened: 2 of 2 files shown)


vidhanio:
i think it would be a good idea to add a ci job to build the nix package as well to avoid future breakages, it wouldn't be too complicated. @crmne would you be fine with it?

```yaml
    - uses: wimpysworld/nothing-but-nix@v9 # optional, provides more storage for nix to use
    - uses: cachix/install-nix-action@v31
    - uses: nix-community/cache-nix-action@v7
      with:
        primary-key: nix-${{ runner.os }}-${{ hashFiles('**/flake.lock') }}
        restore-prefixes-first-match: nix-${{ runner.os }}-
    - run: nix build .#default
```

mixedCase:
@vidhanio Definitely a good idea! Thank you, implemented. We can take it out if undesired, I think the only objectionable thing of that addition is the size of its cache and thus the time it takes to run it cold, where the cache upload took roughly 14 minutes while the actual nix build only took between 1-2 mins. Warm runs naturally ought to be quick.

vidhanio:
@mixedCase yeah the nix github actions cache is kinda something you just test with/without and see if it truly speeds up your ci, in some cases it can make it slower

mixedCase:
Slop machine ran some tests:

 - No cache: 15m 34s                                                                                                                                           
 - Warm cache: 2m 22s                                                                                                                                          
 - Cold cached run: 16m 16s, including the cache upload
 
 So it seems worth it if the storage is not a problem; left it in.

crmne:
This is now in main. Thanks for fixing the Nix build!

PR #117, opened 2026-09-01
Load Now Playing artwork from the disk cache, not the network

iDiegoNL:
## The problem

On macOS, Fastpotify aborts when cover art fails to load.

`media_native.rs` passed `track.art_url` — an `i.scdn.co` URL — to souvlaki as
`cover_url`. souvlaki's macOS backend fetches that itself, synchronously, on a
GCD queue, and then dereferences the result **without checking it**:

```rust
// souvlaki-0.8.3/src/platform/macos/mod.rs:320
unsafe fn load_image_from_url(url: &str) -> (id, CGSize) {
    let url = ns_url(url);
    let image: id = msg_send!(class!(NSImage), alloc);
    let image: id = msg_send!(image, initWithContentsOfURL: url);
    let size: CGSize = msg_send!(image, size);   // <-- nil here
    (image, CGSize::new(size.width, size.height))
}
```

The iOS branch twenty lines above does check for nil. The macOS one does not,
so a struct-returning message to `nil` panics inside an Objective-C callback
that cannot unwind, and the process aborts:

```
thread '<unnamed>' panicked at souvlaki-0.8.3/src/platform/macos/mod.rs:324:24:
null pointer dereference occurred
thread caused non-unwinding panic. aborting.
```

Any track change where the artwork fetch fails takes the whole app down, and
the music with it — an offline laptop, a slow network, a CDN having a bad
minute. It is not `panic = "abort"`; the callback cannot unwind either way.

I hit this through demo mode, where `picsum.photos` was returning 503 — **two
of three launches aborted**. Verified in isolation that the trigger is exactly
what it looks like:

```swift
NSImage(contentsOf: URL(string: "https://picsum.photos/seed/fastpotify0/640/640")!)
// -> nil
```

## The fix

Stop handing the platform a URL. `ArtLoader` has already downloaded the image
and written it atomically (`.part`, then rename), so a file that is present is
a complete, successful response — name that file instead.

- `ArtLoader::cached_file` reports the disk-cache file once it really exists
  and is non-empty.
- `MediaTrack` gains `art_file`. Windows and macOS use it; **Linux is
  untouched** — MPRIS passes the URL along for someone else to fetch and never
  dereferences anything, so it keeps using `art_url`.
- Artwork that has not landed yet is simply left out, and set when it arrives:
  `MediaTrack` derives `PartialEq`, so a track carrying the file differs from
  the same track without it and the existing change detection re-sends.

Two details worth flagging for review:

**The two platforms spell the file differently after `file://`.** macOS parses
the whole string as a URL, so anything URL-significant has to be escaped or the
load fails the same way — I confirmed a `#` in the path (a home directory is
enough to put one there) returns nil and aborts, while the escaped form loads.
Windows trims the prefix and opens the remainder as a plain path, where
escaping would break it. Hence the `cfg` split in `file_url`.

**The controls are synced every frame**, so resolving the file is memoised
rather than stat-ing the disk 60 times a second. That means clearing the
artwork cache has to forget it too, or the remembered path outlives the file
and reintroduces the crash by another door — `Action::ClearArtCache` now does.

## Verification

- **8/8 clean launches** under the condition that aborted 2 of 3 before
  (`picsum.photos` returning 522 throughout).
- 5 new tests: the escaping on each platform, the cache only naming files that
  exist, art arriving counting as a change, and the cleared cache forgetting
  the path. I checked the last one fails without its fix rather than passing
  either way.
- Full checklist from CONTRIBUTING.md green on macOS (arm64): `fmt`, both
  `clippy` runs with `-D warnings`, both `cargo test` runs, doc tests, and
  `rustdoc`. **Not run:** the Jekyll docs build (no docs change).

**Platform coverage, honestly:** built, tested, and run on macOS 26.2 (arm64)
only. Linux and Windows are reasoned about and compile-checked through the
`cfg` split — I have not run either.

## Notes

No docs change: this alters how artwork reaches the OS, not any setting, file,
or user-facing behaviour beyond artwork appearing once downloaded instead of
being fetched a second time by the system. Happy to add a line to
`how-it-connects.md` if you would rather record that the OS no longer makes its
own request for cover art.

The underlying defect is upstream in souvlaki, and this only removes
Fastpotify's way of triggering it. A nil check in `load_image_from_url` is a
few lines and I am happy to send it there too — say the word if you would
rather carry it as a `[patch.crates-io]` entry alongside the `projectm-sys` and
`librespot` ones instead of, or as well as, this.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,6 +1,7 @@
 //! The application: state, event handling, and the actions views ask for.
 
 use std::collections::{HashMap, HashSet};
+use std::path::PathBuf;
 use std::time::{Duration, Instant};
 
 use egui::Color32;
@@ -133,6 +134,10 @@ pub struct App {
     last_settings_save: Instant,
     pub backend: Backend,
     media_controls: Option<MediaService>,
+    /// The artwork the media controls were last given, and the URL it came
+    /// from. Finding the file touches the disk and the controls are synced
+    /// every frame, so the answer is kept until the artwork changes.
+    media_art: Option<(String, PathBuf)>,
     tray: Option<TrayService>,
     pub window_hidden: bool,
     /// The window should close but the process should stay in the tray.
@@ -416,6 +421,7 @@ impl App {
             last_settings_save: Instant::now(),
             backend,
             media_controls,
+            media_art: None,
             tray,
             window_hidden: false,
             hide_intent: false,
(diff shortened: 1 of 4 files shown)


crmne:
This is now in main. Thanks for the fix!

PR #119, opened 2026-09-01
Show the playing song in a panel of its own

iDiegoNL:
The player bar has room for a title, an artist and a heart. Everything else
about the song — who the artist is, who is credited, what plays next — meant
leaving the page you were on.

This puts it beside the page, where the queue and the lyrics already live.
Clicking the artwork in the player bar opens it, the way Spotify's does, or
`N`.

| Before | After |
| --- | --- |
| 1-before  | 2-after-panel

## What it shows

- The artwork at full width, the title and artists, and the same heart the
  player bar carries.
- **About the artist** — their picture, their name, a follower count, and the
  same Follow button the artist page has.
- **Credits** — the credited artists and the album's label.
- **Next in queue** — the next song, and a way through to the rest.

The artist and album land in the caches their own pages read, so opening
either afterwards costs nothing, and the queue behind *Next in queue*
refreshes on the same terms the queue panel already sets.

## Two things Spotify shows that this cannot

Spotify's panel leads with a short looping video and counts an artist's
monthly listeners. Neither is in the Web API, so neither is here:

- **The video** is Spotify Canvas, which lives behind an internal endpoint and
  would need a video decoder besides. The artwork takes its place, which is
  what Spotify itself shows for the many songs with no Canvas.
- **Monthly listeners** are not exposed; the panel counts followers instead,
  and labels it as followers. (Worth noting `/artists/{id}` currently answers
  without `followers` or `genres` for me — the existing artist page has the
  same empty byline — so the line is omitted rather than shown blank.)
- **Credits** name the artists and the label. Writers, producers and mixers
  come from a service the Web API does not expose, and a heading with nothing
  under it seemed worse than a heading with what there is.

I'd rather show less and have it be true than imply a capability that isn't
there.

## Action buttons on hover

The panel stays quiet until the pointer is inside it, then offers a link to
copy, the song's menu, and a button to put the panel away. Whether the song is
saved is worth knowing at a glance, so the heart stays either way (mimics native Spotify).

| Hover | The song's menu |
| --- | --- |
| 3-hover-controls | 4-song-menu |

They fade and slide the last few pixels in rather than snapping into place, and
each reserves the room it lands in so its neighbours hold still while it
travels. The menu counts as being inside the panel — the pointer's trip into it
leaves the panel, and a button that vanished underneath would take the menu
with it.

The menu is the one any row already answers with, rather than a second, thinner
list to keep in sync.

## Behaviour that changed

**Clicking the player-bar artwork used to open the album; it now opens this
panel.** The album is still one click away on the title beside it, which is
where Spotify keeps it too. The chevron that docks the art large still takes
its corner of the cover first. This is the one existing behaviour the branch
changes, and I'm happy to put it back behind a separate button if you'd rather
the cover kept its old job.

## Along the way

Two fixes that are their own commits, and would split out cleanly if you'd
prefer them separately:

- **`Give a menu row with a submenu the icon its neighbours have`** — every row
  in Fastpotify's menus is an icon and a label, except the two that open a
  submenu, which went through egui's plain-label submenu button and started
  their text where the others start their icon. They're now painted like every
  other row, with the arrow that says there's more behind them.
- **`Let one panel own the right-hand side, in one place`** — the rule that only
  one of these panels shows at a time was written out at each of the four
  places that open one. Starting a song's station opened the queue over this
  panel and split the window down the middle, because that path predated the
  third panel. `open_side_panel` now sets all three from the one being opened.

## Light and dark

5-light-theme

## Verification

- Five new tests: the panels' exclusivity, a station not opening the queue over
  another panel, and the headless render test extended to cover the panel with
  a song, with nothing playing, and with the pointer inside it. I checked the
  station one fails without its fix rather than passing either way.
- Full checklist from CONTRIBUTING.md green: `fmt`, both `clippy` runs with
  `-D warnings`, both `cargo test` runs (217 passing), doc tests, and `rustdoc`.
  Not run: the Jekyll build.
- Screenshots are `--demo-shot --demo-show now-playing`; `now-playing` is a new
  demo surface, so the panel is covered by the demo flags like every other one.
- README and the using-fastpotify guide updated.

**Platform coverage, honestly:** written, tested and used on macOS 26.2
(arm64). Nothing here is platform-specific, and Linux and Windows compile, but
I have not run either.

**Note about the demo mode screenshots:** picsum.photos is down, so I had to fall back to generated covers. A fix for that is separate in https://github.com/crmne/fastpotify/pull/117. With the current behaviour on main, Fastpotify crashes when the album art is unreachable (webserver down, no internet, etc).


crmne (changes requested):
Changes requested:

- Keep artwork clicks opening the album or show, and add a separate Info button.
- Put Info and Lyrics in the same right panel as sibling tabs.
- Rebase onto current `main` and split out the unrelated submenu icon change.
- Fix the current CI failures.

The Web API fallback approach is fine.

crmne:
Thanks for this, @iDiegoNL. It has been quiet for two weeks and now conflicts with main, so I'm closing it for now. If you'd like to pick it up again, rebase it, fix the overlapping right-side panels the review found, and include a before-and-after visual comparison (see "Visual reviews" in CONTRIBUTING.md), then reopen. A new panel changes the layout, so it needs my approval of the look before it can merge.

PR #120, opened 2026-09-01
Fully implement folders (write path and increased reliability)

TimotheFCN:
I've promised this PR on X, so here it is.

Since I started it, main got read-only folders (031ef3e), so the tree in the sidebar is already yours. This PR changes two things: folders can be edited, and the shelf says why when they can't be loaded.

You can create folders, rename them, nest them, drag rows in and out, and delete them. Deleting a folder keeps its playlists in Your Library. Changes go back to Spotify, so the desktop and mobile apps pick them up.

Screenshot 2026-09-01 at 04 08 06

## The write path

Writes go through the same private rootlist your read already uses, since the public Web API has no folder endpoint. A change is one protobuf POST to `/playlist/v2/user/<id>/rootlist/changes`, signed with the session's own token.

A change is never sent twice. After the POST, the client re-reads the rootlist and compares it against the exact change it asked for; only a match counts as confirmed.
A 4xx is reported as rejected and the tree stays put.
A timeout or 5xx is treated as unknown, since the change may well have landed, and the readback decides. That keeps a flaky network from duplicating a folder.

Editing raised the bar for reading, so the parsing got strict along the way: unbalanced markers or invalid ids reject the whole snapshot instead of being papered over, because a tree that parsed wrong is a tree that will be written wrong. The last complete rootlist is also cached per account, so the tree shows while the session starts; it is dropped on sign-out and on any read failure, so nobody drags rows around a hierarchy that can't be written. And since a write edits the library of whatever account the session belongs to, editing is only allowed when that account matches the one the Web API is signed in to.

## When folders can't be loaded

Main logs a warning and silently shows the flat list. Here the shelf says what's missing, in place: a line with an **Enable** button when playback is off, a **Retry** button when the read failed, and a different line when playback is on another account. The flat list itself behaves as before, recently-played sort and manual drag order untouched.

Filtering the shelf lists playlists flat, since hidden siblings would make a drop target ambiguous. Nothing polls the endpoint; reads happen on sign-in and after a change.

## Code

- `src/rootlist.rs` is new and holds everything about the format: strict parsing of the URI stream, the row projection the sidebar draws, planning a change into protobuf ops, and checking the result. 22 unit tests cover unbalanced markers, nesting, invalid ids, and each kind of move. `parse_rootlist` and `RootlistEntry` moved there from `src/player.rs`; your paginated `get_rootlist` loop is unchanged and now feeds a snapshot builder.
- `src/player.rs` gains the write.
- `src/ui/sidebar.rs` and `src/ui/dialogs.rs` carry the drop targets and the create, rename, move, and delete dialogs.
- The remembered fold state moves from the session file's `collapsed_folders` to per-account `expanded_folders`, since two accounts don't share folder ids. Old sessions are migrated once the rootlist loads.
- `--demo-show folders` renders a fixture tree for screenshots. The folder paragraph removed from the guide in eb5d041 comes back, rewritten shorter to match the trimmed everyday-use page.

Pins still work: dragging a pinned playlist moves the pin, not the playlist.

## Worth doing later

Folders and playlists only re-read on sign-in, after a change made here, and on a manual refresh. Nothing watches for changes made elsewhere, so a folder created on your phone won't show up here until one of those happens. The window that opens is the real problem: acting on a stale tree means dragging a playlist into a folder that was renamed or deleted a minute ago, and the write goes out against a hierarchy that no longer exists. The readback catches it and the tree snaps back, but the user did work that got undone for reasons they can't see.

Reworking the sync so external changes land quickly would close that. I left it out of this PR to keep the diff reviewable, and because it touches how playlists refresh in general, not just folders.

The session lives inside the player engine, so folder reads and writes reach through it. Pulling the session details out of the engine would make that split cleaner, since the session isn't only a player anymore, but I kept things simple here.
Happy to do it that way instead if you'd prefer.
Same if you'd rather keep the parsing in `player.rs`, and for the wording of the sidebar notice and the docs.


TimotheFCN:
I've reworked this PR now that main reads folders on its own. I started before that landed, but we converged on mostly the same approach. The diff now mostly comes down to the write path and the clearer failure states. The description above is updated to match.

crmne:
Closing because Spotify does not provide a supported folder-write API, and this PR calls a private endpoint. A separate PR for read-only folder caching is welcome. Folder editing can be reconsidered if supported upstream.

PR #121, opened 2026-09-01
Clear stale sidebar context when playing plain tracks

goerll:
## Problem

After starting a playlist from the sidebar, playing a standalone song leaves that playlist green with its speaker icon. Album and playlist plays replace the optimistic context, but a plain track request previously left the old assumed context in place.

## Fix

Represent a plain track list as an optimistic absence of context while Spotify’s playback snapshot catches up. This clears the old sidebar indicator immediately without allowing a stale remote snapshot to restore it.

A regression test covers playing a standalone track while Spotify still reports the previous playlist context.

## Verification

- `cargo fmt --all --check`
- `cargo test --locked --lib` (212 passed)
- `cargo clippy --locked --lib -- -D warnings`

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -302,7 +302,8 @@ pub struct App {
     /// The Premium notice has been shown for this sign-in.
     premium_notice_shown: bool,
     /// The context the interface just started, shown as playing until
-    /// Spotify's own state says the same thing.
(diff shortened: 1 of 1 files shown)


crmne:
This is now in main. Thanks for the fix!

goerll:
Awesome, thank you!

PR #138, opened 2026-09-01
Update installation instructions for Rust and add Fedora dependencies

chinifabio:
## Why

**Build failures on `cargo install`:** Running `cargo install --path .` ignores `Cargo.lock` by default, pulling transitive dependency updates (notably newer versions of `vergen` and `vergen-lib`) that cause compilation errors in `librespot-core`'s build script (`error[E0277]: the trait bound Build: Add is not satisfied`). Adding `--locked` enforces the reproducible dependency tree.

## What changed

- Updated the installation command in `README.md` to `cargo install --path . --locked`.
- Added native system dependency installation instructions for Fedora (dnf) alongside the existing Arch and Ubuntu/Debian instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,12 +91,12 @@ brew install --cask crmne/tap/fastpotify
 Everywhere else, build the single binary with Rust 1.95 or newer:
 
 ```bash
-cargo install --path .
+cargo install --path . --locked
 ```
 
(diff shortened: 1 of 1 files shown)


crmne:
This is now in main. Thanks for improving the docs!

PR #167, opened 2026-09-01
Bring the window back out of the Dock when the Dock asks

### LINKED ISSUE #141: App becomes completely unresponsive after reopening the mini player
### Fastpotify version and how you installed it

0.4.1 from homebrew

### Operating system

macOS

### System details

macOS 26.5.2 m5 MacBook Air 

### Spotify plan

Premium

1. Press play on a song. 
2. Open the Winamp mini player.
3. Minimize the mini player.
4. Reopen the mini player.

**Expected behavior**
The mini player should reopen normally, and the app should remain responsive to user input and macOS menu bar actions.
**Actual behavior**
After reopening the mini player, the entire application becomes unresponsive. The macOS menu bar options for Fastpotify also stop responding.
Interestingly, the music continues playing normally in the background, but the UI itself is completely stuck.

The only way to recover is to force-quit the application using ⌘Q and then launch it again.

Additional notes
This appears to be reproducible when minimizing and then reopening the Winamp mini player.
Winamp skin: Fastpotify

### fastpotify.log

```text
[2026-09-01T15:21:12Z WARN  librespot_connect::state::context] couldn't load context info because: context is not available. type: Default
[2026-09-01T15:21:12Z INFO  fastpotify::sink] audio output: soundcore Q20i
```

iDiegoNL:
Closes #141.

## Why

Minimizing the Winamp mini player was unrecoverable. The window came back
when the Dock icon was clicked, but came back dead: the clock stopped at
whatever second it went down at, the macOS menu bar took clicks and did
nothing with them, and only the music carried on. The only way out was
force-quitting.

## What changed

macOS draws no frames for a minimized window, and every repaint this app
schedules is armed by the frame before it. So the first frame that does not
run is the last one: nothing re-arms, the loop parks in the run loop waiting
for an event, and the queues holding menu picks, tray commands and control
commands are never read again. The window AppKit puts back on screen is the
last frame it drew.

Three things kept it down there, and all three had to go:
1. **`applicationShouldHandleReopen:` only acted when AppKit reported no
   visible windows.** A window sitting in the Dock is reported as *visible*,
   so the one case that needed the handler was the one case it ignored. The
   flag is no longer consulted; asking for a window that is already up costs
   a focus and nothing else.
2. **It left a message without asking anybody to read it.** The `Show` it
   sends can only be read by the frame loop, which by then has stopped, so it
   wakes the app too — the way the status item's own menu items already do.
3. **`Show` then focused a window that was still minimized.** `Focus` does
   not bring a window out of the Dock, so even a frame that was bought would
   have spent it on a no-op. It restores the window first, then focuses it.

Each fix is useless without the others: the first gets the handler to act,
the second buys a frame, the third makes that frame count.

## Verification

**Diagnosis.** Sampling a stuck instance showed the main thread 100% idle in
`__CFRunLoopServiceMachPort` — parked in the run loop, not deadlocked and not
spinning. Logging the delegate then showed it firing with
`has_visible_windows=true` for a window in the Dock, and so doing nothing:

```
17:21:55  reopen handler installed on WinitApplicationDelegate
17:23:07  frame=4020 at 72.6s          ← last frame before the Dock click
17:23:11  applicationShouldHandleReopen visible=true   ← guard bails
17:23:12  frame=4050 at 77.1s          ← one passive frame, then silence
```

**Before**, the mini player is painted but stale — 02:52 on the mini player
against 3:53 in Spotify, the same track:

**After**, the four steps from #141 (play a song, open the mini player,
minimize it, click the Dock icon) return a live window on macOS 26.2.

Commands run, all passing:

```sh
cargo fmt --all --check
cargo test --locked --all-targets                  # 256 passed
cargo test --locked --all-targets --all-features   # 256 passed
```

Two tests, both checked to fail without their fix rather than pass either way:

- `a_dock_click_asks_for_the_window_and_for_a_frame` — fails with *"a window in
  the Dock reports as visible, and was left there"* if the guard returns.
- `showing_a_minimized_window_restores_it_before_focusing_it` — asserts `Show`
  emits `Minimized(false)` before `Focus`.

**Platforms, honestly:** written, reproduced, and tested on macOS 26.2 (arm64)
via the mini player. The change is inside `#[cfg(target_os = "macos")]` code
plus one shared viewport command; Linux and Windows compile and their tests
pass, but I have not run either, and neither has the Dock behaviour this fixes.

- [ ] I updated documentation for user-visible behaviour, settings, files, or network access — **not applicable**: this restores documented behaviour rather than changing it, and adds no setting, file, or network access. Nothing in the README or the guide describes minimizing.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4861,6 +4861,11 @@ impl App {
                     // No window exists; the outer loop creates one.
                     self.wants_show = true;
                 } else {
+                    // A minimized window has to be restored before it can be
+                    // focused: Focus alone leaves it in the Dock, and the
+                    // platform draws no frames while it is down there, so
+                    // nothing arrives to ask a second time.
+                    ctx.send_viewport_cmd(egui::ViewportCommand::Minimized(false));
                     ctx.send_viewport_cmd(egui::ViewportCommand::Focus);
                 }
             }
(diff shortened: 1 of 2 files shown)


crmne:
This is now in main and fixes #141. Thank you!

PR #169, opened 2026-09-01
Prefer a Han fallback face from the reader's region

chagel:
## Why

Follow-up to #10. On macOS, a Simplified Chinese title such as 刘珂矣 or 半壶纱 rendered with its characters at two different sizes and baselines. The han fallback ranks faces by name, and Hiragino Sans, a Japanese font, beat PingFang and Heiti because it calls itself "Sans". It draws the probe character 中 and so passes the coverage check, but it carries no simplified-only characters (刘, 壶, 纱, 纯, 乐, 寻, 风…). Those fell through to the next fallback that had them, Arial Unicode MS, whose em box and baseline differ.

## What changed

A face declares the legacy character sets it covers in the OS/2 table's `ulCodePageRange1` bits (Shift JIS, GB 2312, Wansung, Big5), the same signal Windows font linking reads. The han probe now checks that bit for the locale's region, and a face that does not declare it ranks below every face that does. The rank is an explicit tier in the candidate comparison rather than a numeric penalty, so it cannot drift out of proportion with the naming scores. A wrong-region face still serves when it is all a machine has.

The five region names that were listed in a separate constant now come from the same lookup as the code page bits, so there is one table to extend instead of two.

I checked the bits against the fonts on a current macOS install and they match reality: Hiragino Sans declares only JIS, Hiragino Sans GB declares JIS and GB, Heiti SC/TC declare all five, Apple SD Gothic Neo only Korean, Songti only GB. Noto Sans CJK declares all of them, so on Linux nothing changes and the locale still picks the regional cut by name as before.

## Verification

- macOS 27, `LANG=en_US.UTF-8`: the han fallback moved from `ヒラギノ角ゴシック W4.ttc` (Hiragino Sans) to `Hiragino Sans GB.ttc`; kana still resolve to Hiragino Sans. A demo render of an artist page patched with the titles above shows every character at one size on one baseline. Confirmed in the running app against a real library.
- Arch Linux with Noto Sans CJK: identical before and after.
- `cargo fmt --all --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked --all-targets` (247 passing), and `RUSTDOCFLAGS='-D warnings' cargo doc` are clean. The macOS run used `--no-default-features` because that machine has no CMake; the all-features jobs are left to CI.

- [x] I updated documentation for user-visible behaviour, settings, files, or network access. (No new settings, files, or network access.)

https://claude.ai/code/session_01FoLoTwcjKiy1MxtdEb2RX2

### DIFF
--- a/src/system_fonts.rs
+++ b/src/system_fonts.rs
@@ -9,6 +9,7 @@ use std::path::{Path, PathBuf};
 use std::sync::OnceLock;
 
 use skrifa::MetadataProvider as _;
+use skrifa::raw::TableProvider as _;
 
 /// Registered font name, file bytes, and face index.
 pub struct Fallback {
@@ -58,9 +59,6 @@ const HAN_REGIONS: &[(&str, &str)] = &[
     ("ja", "jp"),
     ("ko", "kr"),
 ];
-
-/// The regional cuts a pan-CJK family can name itself after.
-const HAN_REGION_NAMES: &[&str] = &["sc", "tc", "hk", "jp", "kr"];
 
 /// How deep to walk each font directory. Distributions nest a level or two
 /// (`/usr/share/fonts/truetype/noto`); nothing legitimate goes deeper, and the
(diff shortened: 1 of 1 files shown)


crmne:
This is now in main. Thanks for the thoughtful fix!

PR #171, opened 2026-09-02
feat: Custom scheme support

dim-ghub:
## Why

No way to use custom colour schemes. Users who theme their desktop (e.g. Caelestia) couldn't match Fastpotify.

## What changed

- `ColorScheme` struct: JSON-serialisable palette with hex fields, partial overrides supported
- `ThemeChoice::Custom(String)` variant, keyed by file stem
- Schemes loaded from `~/.config/fastpotify/schemes/*.json`
- Settings UI: separate "Colour schemes" row with toggle buttons
- `Ctrl+Shift+R` / `Cmd+Shift+R` reloads the active scheme live
- "Colour from album art" hidden when a scheme is active
- 6 unit tests, README updated

## Verification

- [x] `cargo fmt`, `clippy`, `test` — all pass (256 tests)
- [x] Linux: created scheme, selected in Settings, toggled between Dark/Custom/Light, reloaded with keybind

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5491,7 +5491,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.3.4",
+ "getrandom 0.4.3",
  "once_cell",
  "rustix 1.1.4",
  "windows-sys 0.61.2",
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ everyday use, and connection details.
   of what you are looking at or listening to. Turn it off in Settings.
 - **Light and dark**, or follow the system. Custom colour schemes are
   supported: drop a JSON file into `~/.config/fastpotify/schemes/` and select
-  it in Settings. Press `Ctrl+Shift+R` / `Cmd+Shift+R` to reload a scheme
+  it in Settings. Press `Ctrl+Shift+R` / `Cmd+Shift+R` or run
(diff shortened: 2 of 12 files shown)


oldirtty:
Nice work!

One gap worth considering before this lands: reload is currently only reachable via the keybind (Ctrl+Shift+R/Cmd+Shift+R), which needs the window focused and manual press it. That works for someone manually tweaking a scheme file, but it breaks the more common real-world flow for desktop theming tools like Noctalia: they regenerate the palette automatically (e.g. on wallpaper change) and run a post-hook to tell every themed app to reload, no window focus need, true hot-reload.

Without a non-interactive reload path, Fastpotify just won't pick up scheme changes from that kind of pipeline; the user has to to hit the keybind every time their theme regenerates, which defeats the point of automatic theming.

A small addition would close that gap: a fastpotify scheme reload (or similar) verb on the existing single-instance control channel, so a post-hook script can do:

```sh
fastpotify scheme reload
```

right after regenerating the palette file, and Fastpotify picks it up immediately, headless, no focus required.

Happy to help wire this up if useful.

dim-ghub:
I experimented with a --watch flag for watching scheme files originally, but you're right a cli command to reload it would probably be best, I can handle that. May take me a bit because I'm currently working on a M3 redesign fork

dim-ghub:
damn it the TCP socket was dropped.

crmne:
Thanks for building this. The interest in theming is clear, but we are not ready to commit to a theme file format and reload API yet. I want to design that once with a future theme browser or store in mind. I am closing this for now rather than keeping you rebasing it.

dim-ghub:
> I want to design that once with a future theme browser or store in mind.

Is the goal to allow for larger themes opposed to simple recolors to support things like #270?

PR #172, opened 2026-09-02
Add podcast playback speed

### LINKED ISSUE #132: Playback speed
### Before requesting

Podcasts are missing the playback speed button. I mostly listen to podcasts at 1.5x.

The Spotify app already has this. The best approach would be to just copy it.

It is a missing feature that the original Spotify app does have. It can be a deal breaker for those who use Spotify to listen to podcasts.

comaeclipse:
## Why

Podcasts have no speed control, which Spotify's own app has (#132). Due to the varying speeds at which people talk, I anticipate this being a highly requested feature.

## What changed

- A small WSOLA time-stretcher (`src/timescale.rs`, no new dependencies, same hand-rolled style as `resample.rs`) runs after the EQ in `vis::Tapped`. It preserves pitch, so a faster voice still sounds like the same person.
- Only podcast episodes are stretched. Music always plays at 1× and bypasses the stage entirely, so songs stay byte-exact.
- A speed pill next to the transport controls cycles Spotify's set: 0.5, 0.8, 1, 1.2, 1.5, 2, 3, 3.5×. It only shows for an episode playing on this computer.
- Live changes ease in over ~60 ms instead of jumping; a track change snaps to the new speed.
- The progress bar's clock runs at the playback speed so it keeps time.
- Persisted as `podcast_speed` in `settings.json` (default 1). README, the guide, and the settings reference are updated.
- `--demo-show podcast` renders the pill for screenshots and tests.

**Known limitation:** while an episode plays here at anything but 1×, other Spotify apps show its position drifting behind until a pause, seek, or the next episode. Spirc only pushes the position to Connect on play/pause/seek/track change and ignores the periodic `PositionChanged`, and `PositionCorrection` never fires when the stream is *ahead* of wall-clock (the player only corrects when behind). Documented in the guide rather than fixed with periodic seeks; the proper fix is a small change in the librespot fork (refresh the Connect position on `PositionChanged` when it drifts by more than 100 ms), which I'm happy to do as a follow-up.

Closes #132.

## Verification

- I've tested it on Windows with real Spotify podcasts at every speed, including live changes mid-episode and episode-to-song boundaries, and it works great.
- `cargo fmt --all --check`, `cargo clippy --locked --all-targets --no-default-features [--features demo] -- -D warnings`, `cargo test --locked --all-targets --no-default-features --features demo` (257 tests), and `RUSTDOCFLAGS='-D warnings' cargo doc --no-deps` all pass. The `--all-features` (libprojectM) variants and the Jekyll build are left to CI; no local toolchain for them.
- Before settling on the window and search sizes I A/B'd the stretcher against the `rodio-wsola` crate on a spoken-word file in a standalone harness; they sound the same, so no dependency.

The player bar with a podcast episode playing at 1.5×

Written with help from Claude Code; I've reviewed every line.


crmne (changes requested):
I previously encouraged this direction, but our current boundary keeps playback capabilities with librespot. We will not maintain a separate time-stretching engine for playback speed, so I’m closing this. Thank you for the implementation and testing.

PR #260, opened 2026-09-02
fix: preserve main window geometry when switching back from Winamp

### LINKED ISSUE #136: Fastpotify using Winamp's window size
### Before requesting

When I swap back to the Fastpotify UI from the Winamp UI, Fastpotify's window is the same size that Winamp's was, and I need to resize it every time.

Make Fastpotify go back to the window size that it was before changing to Winamp, or add an option to have it maximize after swapping back.

Ease of navigation

kveld9:
### Problem

Fixes #136.

Closing the Winamp mini player could overwrite the remembered main-window geometry with the mini window size.

### Solution

- Skip main-window geometry capture during a window switch.
- Preserve the remembered main-window size and position in both switch directions.
- Keep restoration in the existing `App::attach` path.
- Add a regression test for the closing mini-window frame.

### Verification

- Formatting and Clippy
- Linux, macOS, and Windows tests
- Nix package and documentation checks

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -5234,10 +5234,9 @@ impl App {
                 // `main` opens the other kind where each was last.
                 if self.settings.winamp_window {
                     self.winamp.remember_position();
-                } else {
-                    self.session_window_size = self.last_window_size.or(self.session_window_size);
-                    self.session_window_pos = self.last_window_pos.or(self.session_window_pos);
                 }
+                self.session_window_size = self.last_window_size.or(self.session_window_size);
+                self.session_window_pos = self.last_window_pos.or(self.session_window_pos);
                 self.settings.winamp_window = !self.settings.winamp_window;
                 self.settings_dirty = true;
                 self.switch_intent = true;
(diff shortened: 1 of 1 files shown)


crmne (changes requested):
Changes requested:

- Rebase onto current `main`.
- Add a regression test that runs the closing mini-window frame and proves it does not overwrite the saved main-window geometry.
- Save main geometry in both switch directions and skip capture while `switch_intent` is active.
- Keep the existing attach restore path. Remove the duplicate native-options path and hard-coded position bounds unless a failing test proves they are needed.

kveld9:
Rebased and pushed the requested changes:

- Kept `src/main.rs` exactly as it is on `main`, letting `App::attach` handle the restoration path.
- Updated `ToggleWinampWindow` to save `session_window_size` and `session_window_pos` in both switch directions.
- Guarded the geometry capture in `App::frame_ui` with `!self.settings.winamp_window && !self.switch_intent` so the closing frame doesn't overwrite it.
- Added the `closing_winamp_frame_does_not_overwrite_main_window_geometry` regression test using `ctx.run_ui` with a small viewport rect to prove the closing frame leaves the saved main-window geometry intact.

Conflicts are resolved and the new test passes.

crmne (approved):

Issue #122, opened 2026-09-01
Option to start on system startup


gabrieleancora:
### Before requesting

- [x] I searched open and closed issues for this request.
- [x] I read the project boundaries in CONTRIBUTING.md.

### What problem should Fastpotify solve?

Hello,
I always listen to music while using the PC, so I would love to have an option to automatically start fastpotify on system startup.

If the option included the choice to start it minimized, like the native desktop client allows to do, it would be even greater.

Thanks

### What would a good solution look like?

In the settings there would be a dropdown “automatic startup” with options “no/minimized/yes”

### Why does this fit Fastpotify?

It is a feature in the native client and this feature would let me use fastpotify more often than the native client.

Dayflare:
you can just put a shortcut to the fastpotify.exe in the following path: C:\Users\%username%\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup
then it will start when you log in to your windows account.

for a system wide startup for every user you can put the shortcut in the following path: C:\ProgramData\Microsoft\Windows\Start Menu\Programs\Startup

PR #265, opened 2026-09-02
Add startup options

gabrieleancora:
Setting to allow startup on user login. Default is "no"

## Why

Implements  #122 

## What changed

Added option to start Fastptofy on user login. There is a new dropdown in the settings that allows the user to choose to start the application, start minimized or not start at all. 

By default the option is "no". 

## Verification

Tested on Windows all three options. The application is correctly starting up and it is registered on the list of application on automatic startup in the Task Manager.

image

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,8 +119,9 @@ objc2-app-kit = { version = "0.3", default-features = false, features = ["std", 
 objc2-foundation = { version = "0.3", default-features = false, features = ["std", "NSDate", "NSObjCRuntime", "NSRunLoop", "NSString"] }
 
 [target.'cfg(windows)'.dependencies]
-# A hidden window for the media controls to belong to, and message loops.
-windows-sys = { version = "0.60", features = ["Win32_Foundation", "Win32_Graphics_Gdi", "Win32_System_Com", "Win32_System_LibraryLoader", "Win32_System_Threading", "Win32_UI_Shell", "Win32_UI_WindowsAndMessaging"] }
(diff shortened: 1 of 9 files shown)


crmne (changes requested):
Changes requested:

- Rebase onto current `main`.
- Use the Background portal for Flatpak, or disable the setting there.
- Use `SMAppService.mainApp` on macOS, with a deliberate fallback for older supported versions.
- Generate a valid Linux Desktop Entry and test exact escaping.
- Move registration work off the UI thread and save the setting only after success.
- Prevent startup flash and ensure a second autostart launch does not show an existing hidden instance.
- Do not delete startup entries that the user created before this feature.
- Remove the Windows Run entry during uninstall.
- Add focused cross-platform tests and documentation.

crmne:
One more migration blocker: old `settings.json` files have no `automatic_startup` field, so they deserialize to `No`, and startup reconciliation immediately calls `configure(No)`. That unconditionally deletes the generic Linux desktop file, macOS plist, or Windows `Run\\Fastpotify` value even if the user created it manually before this feature existed. Persist an explicit “managed by Fastpotify”/legacy state or verify provenance and the current command before removing anything; absence of a new setting must not delete a user-owned startup entry on upgrade.

crmne:
Thanks for this, @gabrieleancora. It has been quiet for two weeks and now conflicts with main, so I'm closing it for now. If you pick it up again, register the startup entry only after the single-instance check, so a second launch just brings the running window forward, then rebase and reopen. #122 stays open in the meantime.

Issue #175, opened 2026-09-02
Playlist folders only appear sometimes


jorisw:
### Fastpotify version and how you installed it

0.5.0-rc2

Installed by guessing the URL to the DMG

### Operating system

macOS

### System details

macOS 27 Beta 8

### Spotify plan

Premium

### What did you do, and what happened instead?

1. Opened the app
2. Noticed my playlists are flat instead of organized into folders
3. Removed one playlist I didn't need
4. Saw the playlists update into nested folders, as desired
5. Quit and opened the app
6. Got flat playlists again

Possibly fixed by https://github.com/crmne/fastpotify/pull/120

### fastpotify.log

```text
% cat "~/Library/Application Support/me.paolino.fastpotify/fastpotify.log"
cat: ~/Library/Application Support/me.paolino.fastpotify/fastpotify.log: No such file or directory
```

TimotheFCN:
Sadly, folder data is only exposed in librespot through the session API.
With the current design of the app, a session is opened with the local player engine, which causes this reliably issue.

In #120 this is solved by adding a clear warning, as well as a cache so when the session api is unavailable folders stay there (although you cannot make changes to them while the api is unavailable)

So yes we can consider #120 fixes this, but it will still require enabling local playback to work.

jorisw:
> it will still require enabling local playback to work.

What does this mean for the end user, and could the UI in your PR as clearly as possible point to what needs to be done when this is the case?

TimotheFCN:
Basically it means users have to sign in a second time to get folders working.
They only have to do it once.

There is a clear warning at the top of the sidebar in my PR with a quick sign in button :)

crmne:
Fixed on main. Playlist folders now remain visible when local playback is temporarily unavailable. Folder editing remains unsupported.

jorisw:
I've installed v0.6.0, logged out of Fastpotify, logged in to both the shared app and personal app plus enabled playback, but still only see flat playlists.

PR #271, opened 2026-09-03
Read playlists over the streaming session

chagel:
## Why

Opening a playlist could take twenty seconds. The Web API is rate limited per app, the default app is shared by every Fastpotify user, and when its quota runs dry Spotify answers 429 with a long Retry-After that the app rightly honours. A click on a playlist then sits behind that cooldown before anything appears. Debug logs from ordinary sessions show the playlist's own requests taking under half a second once released and 11 to 22 seconds wall clock.

A personal app (the setup in Make It Even Faster) does not reach this case: Development Mode lets it read only the account's own playlists, so everything followed from other people or from Spotify still goes through the shared app.

## What changed

Spotify's own clients read playlists over the connection that streams the audio, which has no per-app quota. The app already uses that session for the rootlist, lyrics, and contributor names. This adds playlist headers and rows, through librespot's typed playlist surface: `Playlist::get_range` from crmne/librespot#2, which main already pins.

Ten commits, one concern each, every one building and passing tests on its own:

1. **Read a playlist over the streaming session.** `src/session_reads.rs`: the header from a zero-length window, pages of rows from windowed reads sliced at the position Spotify reports, and one batched extended-metadata call for track and episode details, shaped as the Web API gives them. `snapshot_id` is the base64 revision, which is what the Web API writes, so the disk cache and the optimistic-write recheck keep matching across both sources. Local files map to the tracks the Web API describes them as, so rows keep their positions. A 404 or 403 on the playlist is the same error the Web API would give; a dropped line is a retry.
2. **Route playlist reads the shared app would get over the session.** `session_serves` in the gateway beside `plan`; the backend offers such a read to the session first, when the engine is up and signed in as the Web API's account, and falls through to the Web API otherwise. Other people's playlists always, the account's own when there is no personal app. Writes, the library list, and search are unchanged.
3. **Keep a playlist's public flag when the session does not carry it.** The header takes it from the library list, when the header arrives and again when the list does, and the edit dialog sends the flag only when its switch was used.
4. **Retry a session page whose song details Spotify did not supply in full.** Each answer in the metadata batch is read with its status. A 404 leaves the row empty, as the Web API shows a song Spotify no longer has. A provider error, any other refusal, bytes that do not parse, or a requested URI missing from the answer fails the page with `Failure::Retry`, so it goes to the Web API instead of being cached with empty rows.
5. **Hand a session page to the Web API when it is short, stalled, or refused.** Session reads get the same 30-second timeout as the Web API client. A window Spotify cut short of the rows that remain is refused rather than paged past. The account gate and the request-to-read mapping are pure functions with tests.
6. **Keep a session page whose song Spotify withholds for legal reasons.** A 451 leaves the row empty, as a 404 does and as the Web API shows it.
7. **Name a playlist's owner where the session does not.** The account's own name and the library list, both from the Web API, fill an owner the session leaves unnamed.
8. **Take a playlist's mosaic cover from the library list.** The session's header carries no picture for a playlist without a cover of its own; the library list does.
9. **Keep a song's known unavailability when the session reads its rows.** The app remembers, by track URI, each song the Web API has said the account cannot play, from every playlist page it reads and from every disk cache it opens, whether or not that cache is still current. Rows that do not say take that mark, on arrival or already on screen, so a greyed-out song stays so wherever it recurs. Rows the Web API never described stay unknown, as the reference page says, and a song it later calls playable is forgotten.

10. **Tidy the session reads after review.** One `Playlist::fill_from` for what a session header takes from the library list, whether the header or the list arrives first. A session error is a retry unless the playlist read itself was refused. The short window is refused before song details are fetched. A sample of a long list's tail, read only for who added the rows, no longer fetches the song details of fifty rows it discards.

The reference pages say which surface serves what, which fields the library list supplies, and how a failed session read falls back.

## Verification

Tested on macOS 27 and on Omarchy (Arch Linux), both with a Premium account and local playback signed in, running with `RUST_LOG=fastpotify=debug,librespot_core=info`.

| Playlist opened | Before, on the shared app | After, with this PR |
| --- | --- | --- |
| Someone else's, or Spotify's | 0.3 to 0.8 s while the shared app is free; 11 to 20 s per request during a cooldown | under 1 s over the session, whatever the shared app's state |
| Own, without a personal app | the same | under 1 s over the session |
| Own, with a personal app | 0.13 to 0.19 s over the personal app | unchanged |
| Shared-app requests per open | 2 to 3 | 0 |
| Session reads handed back to the Web API | | 0 of 137 across the macOS runs |

Web API times are the client's `duration_ms`; session times are the debug log's one-second timestamps from the click to the rows' route line, the same second in four measured opens and the next second in one. The cooldowns Spotify imposed during these runs ran from 1 to 21 s.

- Before: every open of a followed playlist logged `PlaylistMetadata(External) source=shared` followed by `rate limit source=shared wait=19s`, and the requests completed after about 20 s.
- After: other people's playlists, and the account's own without a personal app, log `source=session` for header and rows, with zero `session read failed` fallbacks across a 25-minute session of 53 reads, including after ten minutes idle. With a personal app, own playlists stay on it.
- As asked in review, without a personal app: an own playlist holding the same song twice read over the session with both rows detailed; a rename of that playlist, with the header recheck staying on the session; a forced session failure, from a temporary local patch, falling back to the Web API for the rows while the header stayed on the session; and a song the Web API had greyed out staying greyed out after the session read its list, both with the disk cache still current and with the cache stale and discarded.
- The playlist reads still reaching the shared app are the remembered playing context and last page in the first second after launch, before the engine connects, and the library list; those are tracked in chagel/fastpotify#1.
- `cargo fmt --all --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked --all-targets` (553 lib tests), and `RUSTDOCFLAGS='-D warnings' cargo doc` pass, with `--no-default-features --features demo` on the Mac, which lacks CMake, and the full feature set on Linux. 33 new tests cover the mapping, the batch answer with its statuses, windowing and paging, the failure split, the route matrix, the account gate, the public flag, owner name, cover, and availability handling.

No UI changes, so no screenshots.

https://claude.ai/code/session_01YbKiP9Zuys3nMwXkSpc36a


jorisw:
How does this one relate to #175 ? Will it make it easier/harder/same to fix folders display?

crmne:
Correction: after checking librespot itself, playlist retrieval is already an explicit upstream API. We can use that capability.

Please keep the implementation behind librespot's typed playlist surface. If paging or fields are missing, add them in our librespot fork or upstream first rather than keeping a raw `/playlist/v2/...` client in Fastpotify. Then we can review the routing and fallback on their merits.

This remains separate from #175. Folders use the rootlist, and #175 is Fastpotify failing to preserve the last good tree.

chagel:
> Please keep the implementation behind librespot's typed playlist surface. If paging or fields are missing, add them in our librespot fork or upstream first rather than keeping a raw `/playlist/v2/...` client in Fastpotify. Then we can review the routing and fallback on their merits.


This suggests a reasonable better system design. Let me look into them and separate two PRs under each project concerns.

chagel:
@crmne paging is in the fork as crmne/librespot#2, ready for your review. Once it lands, this PR moves onto `Playlist::get_range` and drops the raw request.

crmne (changes requested):
The typed ranged-playlist API has landed in crmne/librespot#2 (71379a0). Please update the dependency and move these reads onto Playlist::get_range.

The duplicate-item fix is still needed: playables.remove(row.uri()) leaves later occurrences of the same song without metadata. Use a non-destructive lookup and add duplicate-track and duplicate-episode tests.

chagel:
@crmne both changes are in, on four commits rebased onto main. The reads go through `Playlist::get_range` from crmne/librespot#2 and the raw request is gone. Duplicate rows keep their details, with a duplicate-track and a duplicate-episode test. A scenario pass also fixed local-file rows, the public flag when the library list arrives late, and an account check between the session and the Web API. Verified on macOS with a personal app and on Linux without one.

crmne (changes requested):
Playlist rows can lose their unavailable indication with this change, and transient metadata failures can also look like permanently missing songs. playables_of ignores provider_error_status and per-entity status_code; malformed payloads are silently omitted too, so items returns a successful page that can be cached with empty rows. Please distinguish unavailable items from retryable or malformed metadata, return Failure::Retry for the latter, and add partial-batch failure regressions. Also document the missing per-market availability indication in the reference page; the PR body says it is documented, but the current diff does not include it.

chagel:
@crmne addressed in the latest push, rebased onto main.

Metadata answers are now read with their status. Only a per-entity 404 leaves a row without a song, as the Web API shows one Spotify no longer has. A nonzero `provider_error_status`, any other per-entity `status_code`, bytes that do not parse, or a requested URI missing from the response fails the page with `Failure::Retry`, so it goes to the Web API instead of being cached with empty rows. Tests cover the 404 row keeping its neighbours, and each of the four retry causes on a partly failed batch.

The reference pages now carry the per-market availability note the PR body promised, and the fallback paragraph in how-it-connects describes the metadata case.

The librespot pin commit is gone: main already carries `get_range`, so the PR is three commits.

crmne (changes requested):
This changes playlist pages to read through the playback session. The typed ranged reads, duplicate preservation, account matching and incomplete-metadata fallback address the previous blockers.

I want to retain greyed-out availability wherever Fastpotify already knows it. Please preserve known per-account availability when session metadata refreshes a cached track, with a regression for an unavailable song and repeated occurrences. Newly read session-only rows can retain the documented unknown availability. Before merge, please also verify a signed-in playlist with duplicates, an editable playlist and a forced session-read failure that falls back to the Web API.

chagel:
@crmne addressed. The app now remembers, by track URI, each song the Web API has said the account cannot play, from playlist pages and from the disk cache whether or not it is still current, and marks session rows accordingly. A greyed-out song stays greyed out wherever it recurs; rows the Web API never described stay unknown, as documented. Regression covers the unavailable song, its repeat, the stale-cache path, and a song later playable.

Verified live without a personal app: duplicates in an own playlist, a rename with the header recheck on the session, a forced session failure falling back to the Web API, and a greyed-out song staying so with both a current and a stale cache. Full-feature checks pass on Linux. Rebased onto main, with a tidy-up commit on top; the PR body lists the ten commits and a before-and-after table.

crmne (approved):
Playlist pages load through the playback session while keeping known unavailable songs greyed out and preserving the existing editing controls.

I added account-switch and late-cache regressions and fixed the cases they exposed. All contribution checks and native Mac/Windows tests pass. A live playlist matches the Web API in both content and order. The updated CI passed, including Nix. Merged in 47fa97b024d52a809bd47cfcf791fe5e276e4e23. Not released yet.

PR #276, opened 2026-09-03
feat(nix): add darwin appbundle wrapper

ankarhem:
## Why

On nix-darwin the fastpotify binary cannot sit in the Dock or register with Launch Services, so every nix-darwin user has to assemble their own Fastpotify.app wrapper in their config.

## What changed

flake.nix exposes a fastpotify-app package on the two darwin systems, assembled the same way packaging/macos/bundle.sh assembles a release bundle. If you want the release pipeline to use this I can refactor that as well.

See first commit for easier git diff review.

## Verification

- nix fmt
- nix flake check
- nix build .#fastpotify-app
- Tested the fork in my own [nix-config](https://github.com/ankarhem/nix-config/blob/main/modules/general.nix#L49-L86) repo

### DIFF
--- a/docs/_guide/download.md
+++ b/docs/_guide/download.md
@@ -55,37 +55,6 @@ later, allow it in Privacy & Security:
 
 Later launches work with a normal double-click.
 
-### Nix
-
-The repository [flake](https://github.com/crmne/fastpotify) can install the
-app instead of the download above. On macOS,
-`packages.<system>.fastpotify-app` is a `Fastpotify.app` bundle, ad-hoc
-signed and assembled like the release bundle. Nix builds it locally, so it
-is never quarantined and the first-open steps above do not apply.
-
-With nix-darwin, add the flake to your inputs and the bundle appears in
-`/Applications/Nix Apps`:
-
-```nix
-inputs.fastpotify.url = "github:crmne/fastpotify";
-```
-
-```nix
-environment.systemPackages = [
-  inputs.fastpotify.packages."${pkgs.stdenv.hostPlatform.system}".fastpotify-app
-];
-environment.pathsToLink = [ "/Applications" ];
-```
-
-With Home Manager, `home.packages` is enough; its darwin support links app
-bundles into `~/Applications`:
-
-```nix
-home.packages = [
-  inputs.fastpotify.packages."${pkgs.stdenv.hostPlatform.system}".fastpotify-app
-];
-```
-
 ## Windows
 
 The installer adds Fastpotify to the Start menu and needs no administrator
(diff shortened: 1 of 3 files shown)


crmne (changes requested):
Changes requested:

- Copy the executable into the bundle and ad-hoc sign the finished bundle, as `packaging/macos/bundle.sh` does. A symlink to an unsigned store binary does not preserve the launch contract, especially on arm64.
- Document how nix-darwin or Home Manager users install the new output.
- Test opening the resulting app, its Dock presence, Launch Services registration, and a `spotify:` link on the Mac architecture you have. The current verification only proves that the derivation builds.
- Add coverage that checks the assembled bundle and its signature so this does not silently regress.

The separate package output is a useful direction. The release pipeline should keep using its existing signed and notarized path.

ankarhem:
The executable in the store already carries a valid ad-hoc signature (codesign -dv shows Signature=adhoc, and codesign --verify --strict passes on the built .#fastpotify-app), and Launch Services follows the symlink to that binary.

```bash
lrwxr-xr-x 1 root wheel 75 Jan  1  1970 fastpotify -> /nix/store/1jg14vfdnshfp4wib8wbwcspj7aj0iih-fastpotify-0.5.0/bin/fastpotify
/nix/store/1jg14vfdnshfp4wib8wbwcspj7aj0iih-fastpotify-0.5.0/bin/fastpotify: Mach-O 64-bit executable arm64
Executable=/nix/store/1jg14vfdnshfp4wib8wbwcspj7aj0iih-fastpotify-0.5.0/bin/fastpotify
Identifier=fastpotify
Format=Mach-O thin (arm64)
CodeDirectory v=20400 size=203496 flags=0x20002(adhoc,linker-signed) hashes=6356+0 location=embedded
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements=none
Executable=/nix/store/1jg14vfdnshfp4wib8wbwcspj7aj0iih-fastpotify-0.5.0/bin/fastpotify
Identifier=fastpotify
Format=Mach-O thin (arm64)
CodeDirectory v=20400 size=203496 flags=0x20002(adhoc,linker-signed) hashes=6356+0 location=embedded
Signature=adhoc
Info.plist=not bound
```

ankarhem:
- The bundle copies the executable and ad-hoc signs the finished app, like bundle.sh does. The derivation runs codesign --verify --strict on the result, so a broken assembly or signature fails the build instead of shipping.
- Install docs for nix-darwin and Home Manager are in the README and on the download page.
- Verified on aarch64-darwin, installed through nix-darwin: it launches from the Dock, and a spotify: link followed in the browser opens in Fastpotify.

docs/_guide/download.md
```diff
@@ -55,6 +55,37 @@ later, allow it in Privacy & Security:
 
 Later launches work with a normal double-click.
 
+### Nix
```
iamcalledrob: 🚨  Unsolicited drive-by feedback 🚨

These instructions feel overly focussed on nix-darwin, and contain invalid steps for someone who uses NixOS. I'd consider putting the Nix instructions first, and a separate subsection for nix-darwin.

The NixOS instructions being something like:

---

## Nix
Add `crmne/fastpotify` to your flake inputs:
```nix
inputs.fastpotify.url = "github:crmne/fastpotify";
```

Then add the default package to `environment.systemPackages`:
```nix
environment.systemPackages = [
  inputs.fastpotify.packages.${pkgs.stdenv.hostPlatform.system}.default
];
```

### nix-darwin
On macOS, use the `fastpotify-app` package for a `Fastpotify.app` bundle that is built and signed locally.
```nix
environment.systemPackages = [
  inputs.fastpotify.packages.${pkgs.stdenv.hostPlatform.system}.fastpotify-app
];
environment.pathsToLink = [ "/Applications" ];
```

Or use Home Manager to add to `home.packages`, which links app bundles into `~/Applications`
```nix
home.packages = [
  inputs.fastpotify.packages."${pkgs.stdenv.hostPlatform.system}".fastpotify-app
];
```

PR #279, opened 2026-09-03
Add experimental Spotify DJ playback and set controls

derluke:
## Interface changes

Adds a DJ-only button in the main player bar to change sets, plus speaking/changing-set status and an option to open DJ in Spotify. Ordinary Next still skips one song. Local DJ playback turns shuffle off and disables its shuffle control; other playback keeps its existing controls.

Starts Spotify's generated DJ session even when its playlist view is empty, sorted, or filtered. Set changes update the displayed song and queue immediately, preserve manually queued songs, and ignore stale state while playback catches up. Custom text/voice requests remain in Spotify's app.

Depends on [crmne/librespot#1](https://github.com/crmne/librespot/pull/1). All seven engine crates are pinned to its public commit, so the branch builds without local patches or setup scripts.

## Validation

On Linux: formatting, both Clippy configurations with warnings denied, all-target tests with default features (321 tests) and all features (323 tests), documentation tests, strict Rustdoc, and the Jekyll site build pass. Regression tests cover the actual DJ playlist Play button, timeout/retry/reconnection, shuffle normalization, distinct set-change actions, and optimistic queue handling. macOS and Windows have not been tested locally.

## Before merge

- Land the engine change, then move the review dependency pin to the maintainer's merged commit and refresh the Nix vendor hash. The existing hash is not valid for this dependency change; Nix has not been checked locally.
- Review before/after screenshots at representative window sizes in light and dark, and approve the DJ control's visual scope. Demo fixtures: `--demo-show dj`, `dj-speaking`, and `dj-playlist`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2847,7 +2847,7 @@ dependencies = [
 [[package]]
 name = "librespot-audio"
 version = "0.8.0"
-source = "git+https://github.com/crmne/librespot?branch=fastpotify-0.8#5604d0c47673b09b3b18b1f625538ca0f60d63bc"
+source = "git+https://github.com/derluke/librespot?rev=3c79ae742ac1b194a00b28d4b9bfe0c24d94f07a#3c79ae742ac1b194a00b28d4b9bfe0c24d94f07a"
 dependencies = [
  "aes",
  "bytes",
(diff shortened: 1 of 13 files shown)


derluke:
This is a working prototype for discussion. Would you prefer the engine support to land in upstream librespot before considering the Fastpotify integration? Happy to take that work upstream and keep this draft in the meantime.

crmne:
Thanks for putting this together. Spotify DJ is generated for Spotify’s own clients, and neither the Web API nor upstream librespot offers a supported DJ session or controls. Fastpotify does not ship private protocol features solely because they appear in protobufs, so this is outside the project’s boundary. We can reconsider it if Spotify provides an official API or lawful support lands upstream.

PR #289, opened 2026-09-03
Download the cover art the media controls need on macOS

revolutionxk:
## Why

On macOS the Now Playing panel had no album art. Title and artist were fine,
the cover was just blank.

macOS loads the image itself, so `MediaTrack` passes a file path instead of a
URL. But nothing ever downloads that file. `ArtLoader` only fetches what egui
is about to draw, and the player bar draws the 64px `art_small`. The 640px
`art_url` the media controls want never gets requested unless you open the
album page. So `media_art_file` was polling the disk for a file that was never
going to show up.

Linux is fine, MPRIS just forwards the URL.

Before:
image

After:
image

## What changed

New `ArtLoader::prefetch` starts the download for a URL nothing is drawing,
deduped through the existing `entries` map since `sync_media_controls` runs
every frame. `media_art_file` calls it on a cache miss.

Couldn't just point the player bar at the 640px image instead. The Winamp
window draws no art at all, so the controls can't rely on the UI having drawn
something first.

## Verification

macOS 26.6.2, rustc 1.98.0. `cargo fmt --all --check` and
`cargo test --locked --all-targets` pass. Clippy trips on a `type_complexity`
in `src/mac_links.rs` that's already on main, left it alone.

New test in images.rs for the prefetch, and the existing media controls test in
app.rs now checks that a miss actually starts a fetch. By hand: played a song
from a playlist without opening its album, cover shows up now.

Didn't fix one related thing, clearing the art cache deletes the files but not
the loader's in-memory entries, so the playing song stays coverless until the
next track. Can send that separately.


crmne (changes requested):
The missing media artwork fix makes sense. Please limit file prefetch to platforms that use it; Linux MPRIS only needs the URL. Also reject unsupported URI schemes, cover reloading after Clear artwork cache, and document the extra artwork request.

Issue #288, opened 2026-09-03
A click when skipping to another track


Shipi1:
### Fastpotify version and how you installed it

0.5.0, Windows installer

### Operating system

Windows

### System details

Windows 11 24H2

### Spotify plan

Premium

### What did you do, and what happened instead?

1. Started playing a song on this computer.
2. While it was playing, picked a different song.
3. There is a short click right as the new song starts.

It happens whether I press Next or double-click another song in a
playlist. Playing a song from stopped is clean, and letting a track end
into the next one is clean too. It is only when I change song while one
is already playing.

**What I think is happening**: the audio already queued for the sound card
belongs to the old song, and it plays out before the new song starts.
The two waveforms are joined at whatever sample they meet on, and that
join is the click. Skipping also feels like it takes a moment for the
same reason, since that queued audio has to finish first.

### fastpotify.log

```text
[2026-09-03T17:53:53Z WARN  fastpotify::api::client] Spotify rate limit source=shared wait=16s
[2026-09-03T17:53:53Z INFO  fastpotify::api::client] Spotify cooldown source=shared duration_ms=16000
[2026-09-03T17:54:09Z WARN  fastpotify::api::client] Spotify rate limit source=shared wait=1s
[2026-09-03T17:54:09Z INFO  fastpotify::api::client] Spotify cooldown source=shared duration_ms=1000
[2026-09-03T17:54:11Z WARN  fastpotify::api::client] Spotify rate limit source=shared wait=28s
[2026-09-03T17:54:11Z INFO  fastpotify::api::client] Spotify cooldown source=shared duration_ms=28000
[2026-09-03T17:54:39Z WARN  fastpotify::backend] Spotify sign-in verification will retry: Spotify is rate limiting requests; try again in a moment
[2026-09-03T17:54:41Z WARN  fastpotify::api::client] Spotify rate limit source=shared wait=28s
[2026-09-03T17:54:41Z INFO  fastpotify::api::client] Spotify cooldown source=shared duration_ms=28000
[2026-09-03T17:55:09Z WARN  fastpotify::api::client] Spotify rate limit source=shared wait=1s
[2026-09-03T17:55:09Z INFO  fastpotify::api::client] Spotify cooldown source=shared duration_ms=1000
[2026-09-03T17:55:11Z WARN  fastpotify::api::client] Spotify rate limit source=shared wait=28s
[...]

PR #292, opened 2026-09-03
Fade playback in on Play and out on Pause

Shipi1:
Adds a fade to Play and Pause for playback on this computer. Both run 250 ms, so neither is a hard cut.

## Why

Play and Pause were hard edges on local playback: the audio started and
stopped on a step. This is quality of life rather than a fix. Working
properly, you should not notice it at all.

## What changed

Local playback now ramps the output volume instead of switching it.

## How it works
[#288](https://github.com/crmne/fastpotify/issues/288) introduced Envelope, a gain stepped once per frame on rodio's output thread, and TransitionSource, which applies it. This puts a second envelope alongside it:

interrupt — 10 ms, unchanged, smooths a track replaced part way through
transport — 250 ms, new, carries Play and Pause
TransitionSource::next multiplies the two every frame. They stay independent, so a skip landing during a pause rides both without either snapping back. The transport envelope lives on Output and survives the sink swap; the interrupt one is discarded with its sink.

start raises the transport envelope and stop lowers it, before the drain loop that was already there. Since stop already waited for the queue to empty before pausing, the ramp fits inside a wait that was happening anyway, so Pause is no slower than before.

## What is unaffected
- Playback on another device, which never sees the fade
- Gapless playback, since only Play and Pause move the transport envelope
- Seeking
- Track changes, where #288's behaviour is untouched

## Testing and Verification
cargo fmt, cargo clippy -D warnings and cargo test are clean, with 319 passing. Ten tests cover the ramps and the frame count, including that a short queue still fades all the way to silence, that a deep one does not stretch the ramp, and that a chunk rodio discards stops counting as queued.

Listened to a `--release` build: Play, Pause, skip, seek, and pausing
immediately after starting. Everything is working as it should.

Not verified:

- macOS and Linux. Neither built nor heard. `RodioSink` is the default sink
  on all three, so the code path is shared, but I am not claiming more than
  I did.

https://github.com/user-attachments/assets/54f03a58-f037-41c1-adf2-f88a651fab23

https://github.com/user-attachments/assets/2e23a8cd-9520-49cb-8c4f-7a15721e63df

### DIFF
--- a/docs/_guide/getting-started.md
+++ b/docs/_guide/getting-started.md
@@ -68,6 +68,9 @@ You can rename it in Settings.
   from the system tray; reopen it from the tray icon and quit from the tray
   menu or Ctrl+Q. On macOS you can also reopen it from the Dock. Settings can
   turn this off.
+- **Play and Pause fade.** Music played on this computer fade in or out to
+  avoid a hard cut. Skipping between tracks stays gapless and immediate, 
+  and playback on another device is unaffected.
 - **Play buttons show progress.** The button spins until Spotify responds.
 - **Common actions have shortcuts.** Space plays and pauses, Ctrl+F or `/`
   searches, and `Q` opens the queue. Ctrl+/ shows the full list.
--- a/src/sink.rs
+++ b/src/sink.rs
@@ -5,7 +5,7 @@
 //! and reports failures through the UI. Fastpotify can then remain available
 //! as a Connect remote until an output appears.
 
-use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU32, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex, PoisonError, Weak};
 use std::thread;
 use std::time::{Duration, Instant};
@@ -34,6 +34,9 @@ const DRAIN_TIMEOUT: Duration = Duration::from_secs(2);
 
 /// Length of each side of an interrupted-track fade.
 const INTERRUPT_FADE: Duration = Duration::from_millis(10);
+
+/// How long Play takes to come up, and Pause to go down.
+const TRANSPORT_FADE: Duration = Duration::from_millis(250);
 
 /// How often playback looks at which output the system calls its default.
 const DEFAULT_CHECK_INTERVAL: Duration = Duration::from_secs(2);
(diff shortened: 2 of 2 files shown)


Shipi1:
I noticed commit `4c72bf3` that fixes #288 rewrites a good part of `sink.rs`. I have read it, and the sample-clocked `Envelope` is a better mechanism than the one here. I would rather rebuild on it than write over the conflict.


Concretely: this PR ramps gain through `sink.set_volume()` on a wall clock, which can outrun the audio it is fading. That is why it carries `queued: VecDeque<usize>`, `trim_queued`, `fadeable_frames` and `fade_length` : they measure how much sound is left so the ramp can be clamped to it. `Envelope` advances only when a frame is actually pulled, so it cannot outrun its own audio and none of that bookkeeping is needed. Rebuilding on it would delete roughly half this diff and four of its tests, and would drop the "fade length varies with queue depth" caveat in the description.

There is a smaller reason too. `write` now swaps in a fresh `rodio::Sink` on reset, and the deque here tracks the old chunks. Resolving the conflict mechanically would leave it stale.

The one design question I would want your read on: an `Envelope` has a single `frames`, so it cannot be both a 10 ms declick and a 250 ms transport fade. My instinct is two envelopes multiplied in `TransitionSource::next()`, keeping the interrupt and the Play/Pause fade independent so a skip landing mid-fade still behaves. Happy to be told otherwise.

Would you prefer I rebase and rewrite it that way, or leave this as it stands for a first review? 

@crmne

Shipi1:
#288 landed while this was open, and it rewrites the part of sink.rs this touched. Rather than resolve the conflict I rebuilt the fade on Envelope, so this is now an extension of that work instead of a second mechanism beside it.

What is worth your eye:

- **Two envelopes rather than one**. An Envelope has a single level, and the 10 ms interrupt and the 250 ms transport ramp have to sit at different positions at the same time: pause, then skip, and the interrupt must reach zero while the transport holds its place for the next Play. They also have different lifetimes. So TransitionSource multiplies two.
- **Queued and its Drop**. Two monotonic counters, and the Drop impl is the load-bearing part: rodio discards whole sources on sink.stop(), so without settling up there the count would drift upward for good. If there is a simpler way to know how much is queued I would rather use it.
- **Relaxed throughout.** fade_out_over stores step and then target, so the audio thread can briefly see the new target with the old step and move one frame at the wrong rate. Deliberate: nothing is published through these atomics, and the error is smaller than the rounding already accepted in step_over.

crmne (changes requested):
Local Play and Pause gain a transport fade. The first-Play issue is fixed, but pausing after an underrun leaves another hard edge: with zero queued frames, fade_out_over(0) never advances the level to silence. The next start sets the target back to full gain, so the first sample resumes at 1.0 instead of 0.0.

I reproduced this with an open transport envelope followed by fade_out_over(0), fade_in(), and next_gain(). Please settle the transport at silence after stopping, including an empty queue, and add that pause/resume regression.

Shipi1:
Rebased onto main at 0.7.0, no conflicts.

The underrun is as you described. `Envelope::close()` now puts the level at silence outright, and `stop()` calls it after `sink.pause()`, once nothing is being pulled through it. That also covers a drain that gives up at `DRAIN_TIMEOUT` part way down, which left the same edge.

The regression is `a_pause_with_nothing_queued_still_resumes_from_silence`. It fades out over an empty queue, pulls no frame, and asserts the level is still up, then settles and checks that the next Play starts from zero. On the old code it fails at the `silent()` assertion.

PR #300, opened 2026-09-04
Windows Media Player skins

Yukaii:
Wear classic Windows Media Player (.wmz) skins as a transparent chromeless window.

The skin window takes the skin's shape (transparent where the skin leaves nothing, draggable on its shape), sized to the view at 1x to 4x. Layout arithmetic (`treble.left`, bare attributes) and the script's declared numbers (`leftOffset`) settle geometry; button groups hit-test through mapping images with hover/pressed states; sliders, text scrolling, tooltips, and secondary views work. The media pane shows bars/scope or MilkDrop's picture (hidden child renders to shared frames), click to cycle. The skin's close/return button or Ctrl+Esc (Cmd+Esc on macOS) brings the main window back. Skins that will not assemble fall back to the plain player.

Verified: fmt, clippy (default/all-features/no-default-features), tests (410 passed), cargo doc. Swept 186 real skins from the community gallery: 183 load with zero panics across reader, layout, script, and inspector; 3 are corrupt archives (bogus zip offsets) correctly refused. Manually tested Asimov Radio, 9SeriesDefault, Optik, Headspace, Ducky, Classic, and others. Docs (README + guide) included.

Draft while the skin corpus and visual polish settle.


crmne:
This adds a new WMP skin mode and its own layout/script engine. I’m declining that scope for Fastpotify’s current roadmap: the maintenance cost is too large while the existing player and window behavior need attention. Thank you for the work and testing; an experimental fork is a better home for it.

PR #319, opened 2026-09-05
perf(ui): cache table rows and share sorted URI lists with Arc

Cleboost:
## Why

Collection tables rebuild every track row and clone sorted URI lists on each frame. That adds CPU and heap churn on large playlists.

## What changed

- Cache built table rows on `App`, keyed by page, generation, and revisions. At most two tables.
- Drop a cache when its page is evicted, including on `open`, Back, Forward, and cache hits, and on `reset_data`.
- Share sorted URI lists as `Arc<[String]>`.
- Retained size includes nested album, artist, and image strings. The test measures a 500-track collection against names-only, then dropping that page drops the table cache.

User-visible UI: none.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4236,7 +4236,8 @@ impl App {
 
     pub fn open(&mut self, page: Page) {
         if *self.page() == page {
-            self.ensure_loaded(page);
+            self.ensure_loaded(page.clone());
+            self.retain_table_rows(&page);
             return;
         }
         self.history.truncate(self.history_index + 1);
@@ -4246,7 +4247,8 @@ impl App {
         }
         self.history_index = self.history.len() - 1;
         self.show_devices = false;
-        self.ensure_loaded(page);
+        self.ensure_loaded(page.clone());
+        self.retain_table_rows(&page);
     }
 
     /// Hands the app a Spotify link from outside, a canonical URI as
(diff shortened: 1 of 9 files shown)


Cleboost:
This is the first PR in a perf stack. Merge order:

1. #319 (this PR)
2. #321 (art memory, rebased on this)
3. #323 (virtual lists, rebased on #321)

#320 and #322 are independent. #324 is already merged.

Cleboost:
Addressed Copilot review on the latest push:

- rustfmt/clippy findings in \`app.rs\`: fixed in be3f225
- avoid cloning every home preview track: fixed in 63fe699 (only visible rows clone)

crmne (changes requested):
The Arc URI sharing looks useful. The new table cache needs a lifetime and invalidation policy: it retains another full copy of each visited collection, and recreated pages can reuse the same revision numbers. Please cover page eviction and account/data resets, and measure retained memory before and after.

Cleboost:
Addressed the cache lifetime review in 13fc6a1.

The table cache now lives on `App`, keyed by page + generation + revisions, capped at two collections, dropped when the backing page map is gone (the open page is kept while the view temporarily removes that map), and cleared on `reset_data`. Recreated pages that reuse revision 0 miss. `table_rows_retained_bytes()` is the retained-memory measure: a four-track cache is non-zero, then 0 after reset.

crmne (changes requested):
The cache lifetime changes help, but CI now fails in three accessibility tests that still construct RowContext::Uris with Vec. Please update those callers and keep the tests passing. Also wire cache cleanup to actual page eviction: retain_table_rows currently runs only when a new table is stored. The four-track nonzero-size assertion is not a retained-memory comparison; please measure a large collection, including the nested track metadata.

Cleboost:
Addressed in 70e984a. CI is green.

Your last review asked for three things:

1. The three accessibility tests that built `RowContext::Uris` with `Vec` now use `Arc`. Those tests pass.
2. The table cache is dropped on `open`, Back, Forward, and cache hits, not only when a new table is stored. Evicting a page takes its table cache with it.
3. Retained size includes nested album, artist, and image strings. The test measures a 500-track collection against names-only, then dropping that page drops the table cache.

Nothing user-visible.

crmne (approved):
The cache lifetime and accessibility test fixes look good now. The checks pass, including the integration with #322. Ready to merge.

PR #320, opened 2026-09-05
perf(app): evict stale playlist and album pages outside navigation history

Cleboost:
## Why

Playlist, album, artist, and show pages stay in memory after navigation moves on. Long sessions keep every visited page.

## What changed

- LRU page cache. Protected entries are the open page and the playing context, not the whole history.
- Recency updates on `open`, Back, Forward, track-cache reads, and playback cache hits (`cache_track_from_context`, `request_resume_track`). Both recency maps clear on `reset_data`.
- `track_cache` caps at 800 and trims on navigation and the 20s tick.
- Dropped pages also drop their table-row copies.
- Older in-memory pages reload when revisited (from Spotify, or from the on-disk playlist prefix cache when that still matches).

Tests go through real `open`, Back, and paused Next (`step_resume`): after twelve playlists, Back to the first, then a new playlist. The revisited page stays. Stepping to a cached song while paused keeps that song when the cache overflows.

User-visible UI: none.

### DIFF
--- a/docs/_reference/settings-and-files.md
+++ b/docs/_reference/settings-and-files.md
@@ -38,6 +38,11 @@ Successful playlist edits keep that loaded prefix and save it under Spotify's
 new snapshot. Fastpotify reloads the playlist only if the write fails and the
 optimistic edit must be reconciled.
 
+Fastpotify also keeps a small in-memory set of recently opened playlist,
+album, artist, and show pages. Going back to an older page loads it again
+from Spotify, or from the on-disk prefix cache when that still matches the
+snapshot.
+
 The last good playlist folder tree is kept in `session.json`, scoped to the
 account that supplied it. This keeps folders visible when local playback is
 temporarily unavailable. Live session data is still required for edit grants.
--- a/src/app.rs
+++ b/src/app.rs
@@ -1451,6 +1451,8 @@ impl App {
         self.search.results = Loadable::NotLoaded;
         self.search.committed.clear();
         self.table_rows.clear();
+        self.page_used.clear();
+        self.track_used.clear();
     }
 
     /// Drop table-row caches whose pages are gone, and cap what remains.
@@ -4263,7 +4265,7 @@ impl App {
     // ---- navigation ------------------------------------------------------------
 
     pub fn open(&mut self, page: Page) {
-        self.page_used.insert(page.clone(), Instant::now());
+        self.touch_page(&page);
         if *self.page() == page {
             self.ensure_loaded(page.clone());
             self.retain_table_rows(&page);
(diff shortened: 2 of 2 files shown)


Cleboost:
Addressed Copilot review in 7fc6511:

- \`track_cache\` now always trims to 800 entries (sorted keys, playing track kept)
- page eviction uses sorted removable IDs for deterministic behaviour

crmne (changes requested):
The page limits are not enforced when navigation history protects more pages than the limit, and track-cache trimming only runs on navigation. Sorting IDs also is not LRU. Please define and test a bounded cache policy, including data that arrives after navigation and a long browsing session.

Cleboost:
Addressed the bounded cache review in bd9854a.

Protected pages are only the open page and the playing context, not the whole history. Eviction is LRU via last-used time, and track-cache trim also runs on the 20s tick. Tests cover a long history still capping at 12, a page that arrives after navigation, and 900 tracks trimming to 800.

crmne (changes requested):
The limits are bounded now, but recency is not updated on cache hits or Back/Forward navigation. A playlist just revisited through history can still be evicted as the oldest page. Please update those access paths and test eviction through real navigation and track-cache reads, rather than filling the timestamp maps directly. Clear the recency metadata when account data resets too.

Cleboost:
Addressed in e2f2a85. CI is green.

Your last review asked for recency on the real access paths:

- `open`, Back, and Forward update page recency. Track-cache hits update `track_used`.
- `reset_data` clears both recency maps.
- Tests use real navigation, not stuffed timestamp maps: twelve playlists, Back to the first, then a new playlist. The revisited page stays. The oldest unrevisited one goes.

Nothing user-visible.

crmne (changes requested):
Back/Forward is fixed. Playback cache hits still leave old timestamps: stepping through two cached songs while paused can evict the one just previewed. Please update cache_track_from_context and request_resume_track on hits and test through step_resume.

After rebasing, evict_stale_pages also needs to discard #319's table-row copies. Please document that older pages are reloaded when revisited.

Cleboost:
Addressed in 821c04e. Rebased onto main.

- `cache_track_from_context` and `request_resume_track` update recency on hits. Tested through paused Next (`step_resume`).
- `evict_stale_pages` also drops leftover table-row copies.
- Settings docs say older in-memory pages reload when revisited.

quality, docs, and tests are green on Linux, macOS, and Windows.

crmne (changes requested):
Integrated in [b8250c7](https://github.com/crmne/fastpotify/commit/b8250c7), including protection for edits awaiting write or snapshot confirmation. Browsing now releases old cached pages while preserving pending changes. Thanks, @Cleboost. This will be in the release after 0.7.1.

PR #321, opened 2026-09-05
perf(art): touch visible artwork and release JPEG bytes after texturing

Cleboost:
**Merge order:** after #319. This branch is rebased onto #319 so the diff is art-only.

## Why

Cover JPEGs stay in memory after textures exist. Eviction counted a 256 KiB stand-in, which misses the decoded image and the GPU texture.

## What changed

- Touch visible artwork. Drop JPEG bytes once a texture exists.
- Disk reload returns Pending and runs on the runtime. It does not `std::fs::read` on the UI thread.
- The budget is JPEG bytes plus decoded ColorImage plus GPU texture (`2 * width * height * 4`). Eviction forgets the image and the entry.
- The test holds forty 640x640 covers: JPEG-only still fits 64 MiB, decoded plus texture does not. Eviction brings retained size under the budget and drops the oldest scrolled-away cover.

User-visible UI: none.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -4236,7 +4236,8 @@ impl App {
 
     pub fn open(&mut self, page: Page) {
         if *self.page() == page {
-            self.ensure_loaded(page);
+            self.ensure_loaded(page.clone());
+            self.retain_table_rows(&page);
             return;
         }
         self.history.truncate(self.history_index + 1);
@@ -4246,7 +4247,8 @@ impl App {
         }
         self.history_index = self.history.len() - 1;
         self.show_devices = false;
-        self.ensure_loaded(page);
+        self.ensure_loaded(page.clone());
+        self.retain_table_rows(&page);
     }
 
     /// Hands the app a Spotify link from outside, a canonical URI as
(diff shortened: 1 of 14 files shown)


Cleboost:
Merge after #319. This branch is rebased onto #319 so the diff is art-only.

Cleboost:
Addressed Copilot review in 2309823:

- texture-only art entries still count toward the held-byte budget
- disk reload drops the entries lock before reading
- sidebar expanded art and show episode rows now pass \`ArtLoader\`

crmne (changes requested):
The latest load() still calls std::fs::read on the UI thread. Releasing the mutex does not make that read asynchronous. Please schedule cache reloads on the runtime and return Pending. Add cache-clear/reload and eviction tests, justify the texture-memory estimate, and rebase to an art-only diff after #319.

Cleboost:
Addressed the reload and eviction review in 07700a2. Rebased onto #319 so this diff is art-only.

After `release_bytes`, `load()` returns Pending and starts the existing async fetch (disk through `spawn_blocking`). Texture-only entries count as 256x256x4 so the 64 MiB budget still evicts. Tests cover that Pending path, a later Ready from disk, `forget_all`, and eviction with the stand-in size.

crmne (changes requested):
The disk reload is off the UI thread now. Please get #319 green and rebase this stack. release_bytes drops compressed JPEGs, while egui still holds decoded images and textures. The fixed 256 KiB estimate does not account for those allocations. Please use the retained image sizes for the budget and include a before/after memory comparison with large covers and a long scroll.

Cleboost:
Rebased on green #319. Addressed in 23de8c7. CI is green.

Your last review asked to count the memory that actually stays after paint:

- The 256 KiB stand-in is gone. The budget is JPEG bytes plus the decoded image plus the GPU texture.
- After paint, JPEG bytes drop and eviction forgets the texture and the entry.
- Forty 640x640 covers: JPEG-only still fits 64 MiB, decoded plus texture does not. After eviction the oldest scrolled-away cover is gone and retained size is under budget.

Nothing user-visible.

crmne (approved):
The artwork budget now accounts for decoded images and textures, and reloads stay off the UI thread. The checks pass. Ready to merge.

PR #322, opened 2026-09-05
perf(app): snapshot now playing once per frame

Cleboost:
## Why

Playback state was read many times in one frame from a snapshot that could go stale in the same action batch. Idle UI repaint was 4s while paused, even though the Web API poll was already 20s.

## What changed

- Snapshot `now_playing` once per draw. Clear that snapshot before applying actions, so two `TogglePlay` in one batch return to playing (same as main).
- Does not change Web API polling.
- While local playback is paused, idle UI repaint is 20s instead of 4s. Playing still requests 250ms frames.

User-visible UI: none.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -326,6 +326,8 @@ pub struct App {
     #[cfg(feature = "milkdrop")]
     milkdrop_host: Option<crate::milkdrop::host::Host>,
     last_eviction: Instant,
+    /// Playback snapshot for the current frame, built once per redraw.
+    frame_now: Option<NowPlaying>,
     pub sign_in_url: Option<String>,
     /// The verified personal Web API application, when acceleration is ready.
     pub web_app: Option<String>,
@@ -600,6 +602,7 @@ impl App {
             #[cfg(feature = "milkdrop")]
             milkdrop_host: None,
             last_eviction: Instant::now(),
+            frame_now: None,
             sign_in_url: None,
             web_app: None,
             pending_remote_position: None,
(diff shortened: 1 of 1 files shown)


Cleboost:
Addressed Copilot review in 4e3261b:

- refresh \`frame_now\` after \`apply_actions\` before media controls sync
- clear \`frame_now\` at the end of \`frame_ui\` so background work uses live playback state

crmne (changes requested):
Main already polls every 20 seconds during active local playback; this changes the repaint deadline. Please correct the description and measure the saving. The snapshot also needs tests for optimistic actions in the same batch and local/remote handoff before we merge it.

Cleboost:
Addressed the description and snapshot review in 0a6a0a6.

This PR does not change Web API polling. That is already 20s during local playback. The change is the UI idle repaint while local playback is paused: 4s to 20s, about 80% fewer wakeups. Playing still uses 250ms. Tests cover an optimistic pause in the same batch, local to remote handoff, and the 20s idle interval.

crmne (changes requested):
Refreshing after the whole batch still leaves actions reading the old snapshot. I reproduced this with two TogglePlay actions in one apply_actions call: the display ends paused, while the same test on main returns to playing. Please invalidate the drawing snapshot before applying actions, or refresh it between actions, and test that path instead of assigning optimistic_playing directly.

Cleboost:
Addressed in 1955dee. CI is green.

The drawing snapshot is cleared before actions run, not after the whole batch. Two `TogglePlay` in one `apply_actions` call end playing again, same as main. The test drives that path and does not assign `optimistic_playing`.

Nothing user-visible.

PR #324, opened 2026-09-05
perf(vis): reuse FFT scratch buffer in the spectrum analyser

Cleboost:
## Why

The spectrum analyser allocated a fresh FFT scratch buffer every frame.

## What changed

- Keep one reusable scratch buffer inside `Analyser`.
- Zero the buffer before each fill so a short or empty input cannot leak the previous spectrum. A regression compares that result with a fresh analyser.

### DIFF
--- a/src/vis.rs
+++ b/src/vis.rs
@@ -642,6 +642,35 @@ mod tests {
         assert_eq!(scope_shade(15), 4);
     }
 
+    /// A reused analyser that just saw a full frame must match a fresh one
+    /// when the next input is short or empty. Zero-fill stops old samples
+    /// leaking into the spectrum.
+    #[test]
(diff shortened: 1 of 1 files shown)


Cleboost:
Addressed Copilot review in 79095f0:

- zero the FFT scratch buffer before each fill so short sample slices do not leave stale tail values

crmne (changes requested):
Please add a regression that processes a full input followed by short and empty inputs, then compares the result with a fresh analyser. That will protect the zero-fill fix from letting old samples leak into the spectrum.

Cleboost:
Addressed the FFT leak review in bda4014.

The scratch buffer is zeroed before each fill. The new test runs a full input, then short and empty, and compares the spectrum with a fresh analyser so old samples cannot leak.

Issue #104, opened 2026-08-31
Add Japanese font support for Winamp Mini Player


gitHatmos:
### Before requesting

- [x] I searched open and closed issues for this request.
- [x] I read the project boundaries in CONTRIBUTING.md.

### What problem should Fastpotify solve?

In Fastpotify 0.4.0, the Winamp Mini Player was introduced. However, when track titles or artist names contain Japanese characters, they are currently displayed as `?` (question marks)

### What would a good solution look like?

suggestion, a retro bitmap font like Misaki Font might fit well with the Winamp Mini player aesthetic
https://littlelimit.net/misaki.htm

### Why does this fit Fastpotify?

Since the Mini Player uses a pixel font, I chose a Japanese font with a similar aesthetic

crmne:
Fixed in 0.4.1: text the skin's bitmap font cannot say, Japanese included, now scrolls through the marquee in the same pixel face the playlist uses, drawn from the fonts your system already has. Misaki is a lovely suggestion, but borrowing the system's fonts keeps the app small and covers every script at once. https://github.com/crmne/fastpotify/releases/tag/v0.4.1

gitHatmos:
<img width="538" height="136" alt="Image" src="https://github.com/user-attachments/assets/327b95f9-0e2d-4a55-926f-a499fa36011d" />

Thank you for the quick fix!

I checked it right away. While Japanese titles and artist names are now displayed, the characters appear squished/crushed (I'm not sure if it's due to font size or scaling issues)

crmne:
Good eye, that was the font's own padding being squeezed into the bar along with the glyphs. Fixed on main: the drawing now scales the ink itself to the bar's height, so the characters fill it. It will be in the next release.

gitHatmos:
Also, I wanted to share what seems to be a related bug

In the main window, I noticed that when an album title contains a full-width tilde (`～`), the text seems to prioritize rendering in the pixel font. After checking further, it appears likely that all full-width special characters are being displayed in the pixel font as well.
(Using Fastpotify v0.4.1, Windows 11 25H2)

This is using full-width tilde in album title
Album : https://open.spotify.com/album/3pkS4JGRAhXDI9kc9OouPH
Title : ＊〜アスタリスク〜
<img width="800" height="242" alt="Image" src="https://github.com/user-attachments/assets/cf0e8136-d23a-4f4d-aa32-87ae93d66cf4" />

[...]

PR #332, opened 2026-09-05
Fold full-width punctuation onto the skin font's own cells

kevin9327:
## Why

A title punctuated with full-width characters loses the skin's font entirely, not just those
characters.

`covered` decides per character whether `text.bmp` has a real cell for it, and both callers ask the
question of the whole string:

```rust
if text.chars().all(font::covered) { …skin font… } else { …borrowed face… }
```

`fold` already reduces a character to one the font has: lower case takes the capital, an accent is
dropped, a curly quote takes the straight one, `~` takes the dash. Full-width forms are missing from
it, so `＊`, `～`, `（`, `）` and the rest all reach `cell` unchanged, land on the question mark, and
`covered` says no. One of them anywhere in a title sends the entire line to a borrowed face.

They are not rare in the titles this affects. A Japanese or Korean release writes `（2026）`,
`ＲＥＭＩＸ`, `～`; the full-width forms are the normal punctuation there.

The font's own note says accented Latin letters "lose their accents rather than turning into
question marks". A full-width star is a star and a full-width tilde is a tilde, so the same rule
covers them, and here it decides a whole line rather than one glyph.

## What changed

`fold` maps U+FF01 to U+FF5E onto the ASCII range they mirror, which is a fixed offset of 0xFEE0,
and the ideographic space U+3000 onto a space. The existing arms then apply as they already do, so a
full-width lower-case letter still takes its capital and the full-width tilde still takes the dash
the ASCII one takes.

`covered` asks its question after folding. It used to compare the raw character against the question
mark, so `？` folded onto a cell the font really has and was still reported as uncovered. Now
anything that folds onto the question mark counts as covered by it, which is what the fold means.

Nothing else moves: a character outside those ranges reaches the same match it did before, and kana,
Han and hangul are still uncovered and still fall back, which is correct — the font has no cell for
them.

## Before and after

The mini player, built-in skin, playing a title whose only non-ASCII is full-width punctuation
(`--demo --demo-show winamp`, with the demo track renamed to `Rosewood（2026）～REMIX` for the shot):

Before — one full-width character sends the line to the borrowed face:

before

After — the same title in the skin's own cells:

after

## Tests

- `full_width_forms_fold_to_the_shapes_they_are` — `＊` is the star cell, `（` and `）` the bracket
  cells, `Ａ` and `ａ` the capital, `０` the digit, U+3000 the blank, `～` the dash, `？` the question
  mark and `covered('？')` is true. Before this change the star assertion fails with `left: (2, 3)`,
  the question mark, against `right: (2, 4)`, the star; before the `covered` change the last
  assertion fails on its own.
- `a_title_punctuated_full_width_keeps_the_skin_font` — the whole-string question both callers ask
  now answers yes for `（2026）　REMIX～`, and still answers no for kana, so the fallback is not
  weakened.

## How I tested

`cargo test --locked --all-targets` (326 passed, up from 324 by the two new tests) and
vcpkg for MilkDrop; nothing here touches that feature. The captures above are from a
`--features demo` build. No platform-specific code.

## On #104

I overstated the connection in the first description, and have corrected it above. #104's follow-up
is about font fallback in the main window, which this change does not touch. What I took from that
thread is only the example: `＊〜アスタリスク` is the kind of title people were reporting, and it
still falls back here because of the kana. This change is about the neighbouring case — a title
whose only non-ASCII is full-width punctuation, which had no reason to lose the skin font and did.

### DIFF
--- a/src/skin/font.rs
+++ b/src/skin/font.rs
@@ -34,7 +34,9 @@ pub fn cell(character: char) -> (u32, u32) {
 /// Whether the font has a real cell for the character, rather than the
 /// question mark it falls back to.
 pub fn covered(character: char) -> bool {
-    character == '?' || cell(character) != cell('?')
+    // Asked after folding, so a character that folds onto the question mark
+    // is covered by it: the full-width one is the same question mark.
+    fold(character) == '?' || cell(character) != cell('?')
 }
 
 /// The sprite for a character.
@@ -120,6 +122,10 @@ mod tests {
         assert_eq!(cell('\u{3000}'), cell(' '));
         // The full-width tilde takes the dash the ASCII one already took.
         assert_eq!(cell('\u{ff5e}'), cell('-'));
+        // The full-width question mark is the question mark, so it is a cell
+        // the font has rather than the fallback for one it does not.
+        assert_eq!(cell('\u{ff1f}'), cell('?'));
+        assert!(covered('\u{ff1f}'));
     }
 
     #[test]
--- a/src/skin/font.rs
+++ b/src/skin/font.rs
@@ -45,6 +45,17 @@ pub fn glyph(character: char) -> Sprite {
 
 /// Reduces a character to one the font has a cell for.
 fn fold(character: char) -> char {
+    // Full-width forms are the ASCII punctuation, digits and letters again,
+    // shifted up by 0xFEE0, and a Japanese or Korean release writes a title
+    // with them: a star is U+FF0A, a tilde U+FF5E, the brackets U+FF08 and
+    // U+FF09. Folding them is the same move as dropping an accent, and it
+    // matters more, because one uncovered character sends the whole line to
+    // a borrowed face instead of the skin's own font.
+    let character = match character {
+        '\u{ff01}'..='\u{ff5e}' => ((character as u32 - 0xfee0) as u8) as char,
+        '\u{3000}' => ' ',
+        other => other,
+    };
     match character {
         'a'..='z' => character.to_ascii_uppercase(),
         ' ' | '\t' | '\u{a0}' => ' ',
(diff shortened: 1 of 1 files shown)


crmne (changes requested):
This changes the mini player font for titles containing full-width forms. Please add before/after captures. Also, covered('？') still returns false even though fold maps it to the supported question mark. Please cover that case in the mapping and tests.

The follow-up in #104 concerns font fallback in the main window, which this change does not touch. Please correct that connection in the description.

kevin9327:
All three addressed.

**`covered('？')`** — you are right, and it was the same bug one level up. `covered` compared the raw character against `'?'`, so a character that folded onto a cell the font really has was still reported as uncovered. It now asks after folding:

```rust
pub fn covered(character: char) -> bool {
    fold(character) == '?' || cell(character) != cell('?')
}
```

`cell('？') == cell('?')` already held from the fold, so the new assertions are `cell` and `covered` both, and `covered` failed on its own before this commit.

**Captures** — added to the description: the mini player on the built-in skin, `--demo --demo-show winamp`, demo track renamed to `Rosewood（2026）～REMIX` for the shot. Before is the borrowed face mid-scroll, after is `ROSEWOOD(2026)-REMIX` in the skin's cells.

**#104** — corrected. I had claimed this was the second half of that report; it is not. The follow-up there is main-window font fallback and this does not touch it. All I took from the thread is the example, and `＊〜アスタリスク` still falls back here because of the kana. The description now says so.

326 tests pass, fmt and clippy clean.

crmne (approved):
Mini-player titles containing full-width punctuation now keep the skin font. The question-mark regression, before/after captures, and corrected description address the review. All CONTRIBUTING checks pass locally on Linux, and the three-platform CI is green.

PR #333, opened 2026-09-05
Replace clap with usage-rs

seuros:
Much cleaner and faster implementation of clap by @jdx .


crmne (changes requested):
I’m closing this parser replacement because it has no demonstrated size or startup benefit, and the Clippy and Nix failures remain. I would reconsider a focused proposal with before/after measurements and coverage preserving the existing CLI behavior.

PR #335, opened 2026-09-06
feat: follow the Linux desktop palette in system theme

Hank-dev:
## Summary

- On Linux, Appearance, Follow system now uses the active Omarchy desktop palette (`~/.config/omarchy/current/theme/colors.toml`) for window, text, and accent colours.
- Explicit Light and Dark still use Fastpotify's built-in palettes.
- Surfaces are mixed from background and foreground so the UI stays readable instead of using raw terminal ANSI blacks.

## Motivation

Follow system previously only selected light or dark. On Omarchy (and similar Linux setups that publish that colour file), the rest of the desktop already has a full palette. Fastpotify stayed on Spotify green over a generic dark grey.

## Changes

- `Palette::from_system` is Linux-only for the file-based palette. Other platforms still follow OS light/dark.
- The colour file is re-read on the UI thread when its mtime changes. There is no leftover watcher thread after leaving Follow system.
- Incomplete writes keep the last valid palette. Deleting the file falls back to the built-in light or dark colours.
- The parser accepts comments, single or double quotes, 3/6/8-digit hex, and uppercase values.
- README and settings docs describe the Linux palette file.

## Test Plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --locked --all-targets -- -D warnings`
- [x] `cargo test --locked --lib` (332 passed)
- [x] `cargo test --locked --test system_palette_reload`
- [x] Live check on Omarchy: Follow system picks up the current violet palette
- [ ] CI on macOS and Windows (compile-only coverage for the `cfg` split)
- [ ] `cargo clippy --locked --all-targets --all-features -- -D warnings` and the remaining CONTRIBUTING checks

## Screenshots / Examples

This is a colour change only: layout, fonts, and controls are unchanged. Live Linux screenshot after the change: system violet accent and deep purple-black surfaces instead of Spotify green.

Demo mode still uses the built-in palettes unless that Omarchy file is present in the test environment.

## Notes for Reviewers

This is not a general GTK or portal theming engine. It only reads Omarchy's published `colors.toml` on Linux. If that path is too specific, it can be narrowed or made a documented Linux extra without changing the Light/Dark behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,9 @@ everyday use, and connection details.
   starts.
 - **Album-art colour.** Pages and the player bar take a tint from the cover
   of what you are looking at or listening to. Turn it off in Settings.
-- **Light and dark**, or follow the system.
+- **Light and dark**, or follow the system. On Linux, follow system also
+  uses the active Omarchy desktop palette when
+  `~/.config/omarchy/current/theme/colors.toml` is present.
 - **Winamp mini player.** `Ctrl+M` opens a small player for classic `.wsz`
   skins, drawn at 1x to 4x scale. It includes a spectrum analyser, playlist,
   and equalizer. Drop a skin from the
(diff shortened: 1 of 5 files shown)


crmne (changes requested):
This changes Follow system into an Omarchy palette reader. I’m keeping Follow system’s existing light/dark meaning and choosing the generic JSON theme direction in #392. Closing this competing integration; any future theme-file loading must also stay off the UI thread.

PR #337, opened 2026-09-06
Read an unpacked skin from the folder it was unpacked into

kevin9327:
## Why

A skin that reads perfectly well as a `.wsz` stops reading the moment somebody unpacks it, if its
files sit under a folder inside the archive.

The two readers disagree about exactly one thing. `from_archive` discards the folder part of every
name:

```rust
pub fn base_name(&self) -> &str {
    self.name.rsplit(['/', '\']).next().unwrap_or("")
}
```

which its own doc calls out: "File names are matched without regard to case or folder". `from_dir`
reads one level and nothing else, so `Some Skin/main.bmp` on disk is not `main.bmp` and the reader is
told

```
no skin bitmaps were found inside
```

about a folder that plainly has a skin in it. The stripping in the archive reader is the evidence
that these names are common: it would not be there otherwise, and unpacking is what turns those
names into folders on disk.

## What changed

`from_dir` looks into the folders inside as well, matching names the way the archive reader already
does: lower cased, folder discarded.

A folder's own files go in before its subfolders are read, and a name already present is not
replaced, so the shallower copy wins — the folder the reader pointed at beats anything nested under
it. A flat folder, which is what every skin that works today looks like, is read in exactly the
order it was before and gets exactly the same files.

Links are not followed: a file type from `read_dir` reports a link as a link rather than as a
directory, so a link back to a parent is skipped rather than walked. `MAX_SKIN_DEPTH` is the cheap
second answer to the same question, for a file system where that is not true.

## Tests

`a_folder_skin_is_read_from_the_folder_it_was_unpacked_into` builds the case on disk: a folder
holding `Some Skin/MAIN.BMP` and `Some Skin/pledit.txt`, then asserts the main sprite's pixel and the
playlist colour. Before this change it fails with

```
called `Result::unwrap()` on an `Err` value: Empty
```

`a_folder_of_bitmaps_is_a_skin_too` is unchanged and still passes, which is the flat case.

## How I tested

`cargo test --locked --all-targets` (347 passed, up from 346 by the new test) and

No user-visible setting or network access changes. It reads more of a folder the reader already
chose, so no documentation change.

### DIFF
--- a/src/skin/mod.rs
+++ b/src/skin/mod.rs
@@ -134,15 +134,18 @@ impl Skin {
     }
 
     /// Reads an unpacked skin: a folder with the bitmaps in it.
+    ///
+    /// Folders inside it are looked into as well. A `.wsz` names its files
+    /// under a folder often enough that the archive reader discards one, so
+    /// a skin that reads perfectly well as an archive stopped reading the
+    /// moment somebody unpacked it: the bitmaps were one level down and the
+    /// answer was "no skin bitmaps were found inside".
+    ///
+    /// The shallower copy of a repeated name wins, which is the folder the
+    /// reader pointed at winning over anything nested beneath it.
     pub fn from_dir(name: impl Into<String>, dir: &Path) -> Result<Self, SkinError> {
         let mut files = Files::new();
-        for entry in std::fs::read_dir(dir)? {
-            let entry = entry?;
-            let file_name = entry.file_name().to_string_lossy().to_ascii_lowercase();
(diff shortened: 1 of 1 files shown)


crmne (changes requested):
Integrated in [bc4f045](https://github.com/crmne/fastpotify/commit/bc4f045), with the depth-precedence correction and regression coverage. Unpacked skins now load from nested folders, using the closest copy when file names repeat. Thanks, @kevin9327. This will be in the release after 0.7.1.

PR #338, opened 2026-09-06
Keep a folder name whose percent begins no escape

kevin9327:
A playlist folder name with an unescaped percent can panic while Fastpotify reads the playlist tree. For example, `100% 안녕` makes the decoder end a string slice inside a UTF-8 character. The correctly encoded `100%25 안녕` already works.

Read the two candidate hex digits as bytes. Valid escapes decode as before; an invalid escape keeps its percent sign and continues without slicing UTF-8 text.

The regression covers both the malformed name and the correctly encoded compatibility case. The malformed case reproduces the panic on main; the encoded case passes before and after.

The contributor tested on Windows 11 without default features. All checks from CONTRIBUTING.md pass locally on Linux, including default and all-feature Clippy/tests, documentation tests, Rustdoc, and the website build.

### DIFF
--- a/src/player.rs
+++ b/src/player.rs
@@ -910,16 +910,25 @@ fn decode_folder_name(encoded: &str) -> String {
                 out.push(b' ');
                 i += 1;
             }
-            b'%' if i + 2 < bytes.len() => match u8::from_str_radix(&encoded[i + 1..i + 3], 16) {
-                Ok(byte) => {
-                    out.push(byte);
-                    i += 3;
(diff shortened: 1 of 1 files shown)


crmne (approved):
Malformed percent escapes in playlist folder names now stay readable without a UTF-8 slicing panic. The valid encoded case is preserved. I reproduced the failure on main, reviewed the fix, and ran all CONTRIBUTING.md checks on Linux; all six GitHub checks also pass.

PR #339, opened 2026-09-06
Cut a featuring where it starts, not where the lower-cased copy says

kevin9327:
## Why

The featuring cut is looked for in a lower-cased copy of the title and then made in the title
itself:

```rust
let lower = text.to_lowercase();
...
let before = lower[..start].trim_end_matches(['-', '(']).trim_end();
let at = before.len();          // an offset into `lower`
...
Some(at) => text[..at].trim().to_string(),   // used as an offset into `text`
```

Lower casing is not length-preserving. Turkish `İ` (U+0130) lower cases to two characters and one
byte more; capital sharp s `ẞ` (U+1E9E) loses one. Past either of them the two strings no longer
agree about where anything is, so the offset means somewhere else in the title.

Too far, and letters that should have gone are kept:

```
clean_title("İİ feat. Someone")
  left:  "İİ f"
  right: "İİ"
```

Too short, and it lands inside a character, which is not a wrong answer but a panic:

```
clean_title("ẞ café feat. Someone")
thread panicked at src/lyrics.rs:426:25:
end byte index 8 is not a char boundary; it is inside 'é' (bytes 7..9 of string)
```

Both are reached by playing a track and opening the lyrics: `clean_title` and `clean_artist` are
what the LRCLIB lookup is built from, and both run `strip_featuring`. `İ` is an ordinary letter in
Turkish and Azerbaijani, and a title of the form `Şarkı feat. Biri` is the normal way a
collaboration is written there.

## What changed

`strip_featuring` builds the lower-cased copy itself, keeping a table beside it that carries every
byte offset in the copy back to the same place in `text`. The search is unchanged and still runs in
the copy; only the final cut is translated:

```rust
Some(at) => text[..starts[at]].trim().to_string(),
```

`starts` is built by walking `text.char_indices()` and pushing the source character's offset once
per byte the lowered form occupies, so a character that lowers to two still resolves to the place it
came from. `starts.push(text.len())` covers a cut at the very end.

For a title where lower casing is length-preserving -- which is every ASCII title and the
overwhelming majority of the rest -- `starts[at] == at`, so nothing about the existing behaviour
moves.

## Tests

`a_featuring_is_cut_where_it_starts_whatever_the_case_costs`:

- `İİ feat. Someone` cuts to `İİ`, and `İzel feat. Someone` (a real Turkish singer) to `İzel`
- `ẞ café feat. Someone` cuts to `ẞ café` instead of panicking
- `Song feat. Someone`, `Song ft. Someone` and `Artist featuring Other` are unchanged, which is the
  check that the ASCII path did not move

On `main` the first assertion fails with `left: "İİ f"` and the third panics.

## How I tested

`cargo test --locked --all-targets` (347 passed, up from 346 by the new test) and

### DIFF
--- a/src/lyrics.rs
+++ b/src/lyrics.rs
@@ -401,7 +401,25 @@ pub fn clean_title(title: &str) -> String {
 
 /// Everything from a standalone "feat", "ft", or "featuring" on.
 fn strip_featuring(text: &str) -> String {
-    let lower = text.to_lowercase();
+    // The marker is looked for in a lower-cased copy and the cut is made in
+    // `text`, and lower casing does not preserve length: Turkish 'İ' becomes
+    // two characters and one byte longer, capital sharp s one byte shorter.
+    // An offset measured in the copy therefore means somewhere else in the
+    // title -- letters kept that should have gone, or, when it lands inside a
+    // character, a panic. `starts` carries every offset in the copy back to
+    // the same place in `text`, so no offset is ever used in the string it
+    // was not measured in.
+    let mut lower = String::with_capacity(text.len());
+    let mut starts: Vec<usize> = Vec::with_capacity(text.len() + 1);
+    for (at, character) in text.char_indices() {
+        for lowered in character.to_lowercase() {
+            for _ in 0..lowered.len_utf8() {
+                starts.push(at);
+            }
+            lower.push(lowered);
+        }
+    }
+    starts.push(text.len());
     let mut cut = None;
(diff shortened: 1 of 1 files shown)


crmne (approved):
Lyrics lookup now cuts featuring credits at the correct character boundary for Unicode titles and artists. I reproduced both the extra-letter result and the panic on main; the new regression covers both. All CONTRIBUTING.md checks pass locally on Linux with the other merged fixes, and all six GitHub checks pass.

PR #340, opened 2026-09-06
Fit the last row to the column even when one word is longer than it

kevin9327:
## Why

For right-to-left text, `layout` wraps the rows itself in `break_rows` and hands them to epaint with no wrap width (`f32::INFINITY`). So if a row comes back wider than the column, nothing downstream catches it, and the row is drawn past its cell.

Two cases produced such a row:

- The first word of a row is taken without being measured, because a row has to hold something. A single word longer than the column therefore came back whole, with no ellipsis.
- Once a row was cut, the width left was worked out by subtracting the widths of isolated letters. That is wrong for joining scripts. At 18 px, `لالالالالالا` cut for a 25 px column was drawn about 49 px wide with the fonts the app installs on Windows (43.53 px in the review).

This happens with an Arabic, Hebrew or Persian title in a narrow list cell or sidebar. `ellipsized` always passes `Some(ELLIPSIS)`, and list rows use one row.

## What changed

Only the last row, which is the row the ellipsis belongs to, and only when it was cut or does not fit.

- **Shaped measurement.** Each candidate is measured the way `layout` draws it: the row plus the mark, in display order through `display_line`, laid out whole. The existing loop that gives up whole words uses the same measurement.
- **Cutting inside a word.** When one word is still too wide, the cut is found by bisecting over allowed cut points. A prefix is kept only after it has been measured inside the column. Cuts never separate a mark (bidi class NSM or BN) from the letter before it, and never fall right after a ZWJ.

A middle row holding a word longer than the column is left as it is, because fitting it would drop letters with no mark to show it. Breaking a word across rows would be a separate change.

## Tests

`a_shaped_word_longer_than_the_column_is_drawn_inside_it` goes through the real path. It calls `theme::install`, runs a pass so the fonts apply, and then calls `layout` at 18 px for an Arabic word (`لالالالالالا`) and a Hebrew word (`והתקשרויותיהם`). Each word is laid out in columns of 0.4, 0.6 and 0.85 of its own shaped width (never narrower than the mark), on one and on two rows. The test asserts that the drawn row's width is within the column and that the row carries the mark. It also checks that, given room, the word stays whole and unmarked. Columns are fractions of the measured width, not fixed pixels, because runner fonts differ. With fixed 25/40/60 px columns, the Arabic word fit one of them on macOS and Windows CI.

```
لالالالالالا in a 25.1375 px column is drawn as …لالال, 49 px wide
```

The character-count tests `a_word_longer_than_the_column_still_fits_the_column` and `rows_that_already_fit_are_left_alone` keep the Latin-script behavior pinned: `abcdefghij` in 5 becomes `abcd…`, rows that fit gain no mark, and the space cut still gives `ab…`.

## How I tested

- Windows 11, toolchain 1.98.0 from `rust-toolchain.toml`, with the fonts the app installs (Noto Sans Arabic).
- All 11 `bidi` tests pass. `cargo fmt --all --check` and `cargo clippy --lib --tests -- -D warnings` are clean with `--no-default-features --features demo` (MilkDrop needs vcpkg GLEW, which this machine does not have).
- CI is green on 8e1b303: quality, docs, Nix package, and tests on Ubuntu, macOS and Windows.
- Before/after captures of long Arabic and Hebrew titles, at 1240x800 and 760x800 in dark and light themes, are in [the reply to the review](https://github.com/crmne/fastpotify/pull/340#issuecomment-5653691360).

### DIFF
--- a/src/bidi.rs
+++ b/src/bidi.rs
@@ -193,16 +193,33 @@ fn break_rows(
         }
     }
     rows.push(row);
-    if cut && let Some(mark) = overflow {
+    // The last row can be wider than the column even when nothing was cut:
+    // the first word of a row is taken whatever it measures, because a row
+    // has to hold something, and a word longer than the column has nowhere
+    // to break. `layout` hands the finished rows to epaint with no wrap
+    // width of its own -- the fitting is meant to have happened here -- so a
+    // row left too wide is drawn straight past the edge it was given.
+    if let Some(mark) = overflow {
         let mark_width = width(mark.encode_utf8(&mut [0; 4]));
         let last = rows.last_mut().expect("one row at least");
-        while last.1 + mark_width > wrap_width
-            && let Some(at) = last.0.rfind(' ')
-        {
-            last.1 -= width(&last.0[at + 1..]) + space;
-            last.0.truncate(at);
+        if cut || last.1 > wrap_width {
+            while last.1 + mark_width > wrap_width
+                && let Some(at) = last.0.rfind(' ')
+            {
+                last.1 -= width(&last.0[at + 1..]) + space;
+                last.0.truncate(at);
+            }
+            // A single word with no space left to give up: characters go
+            // instead, which is the only way the ellipsis can mean anything.
+            while last.1 + mark_width > wrap_width
+                && let Some(character) = last.0.chars().next_back()
+            {
+                let at = last.0.len() - character.len_utf8();
+                last.1 -= width(&last.0[at..]);
+                last.0.truncate(at);
+            }
+            last.0.push(mark);
         }
-        last.0.push(mark);
     }
     rows.into_iter().map(|(text, _)| text).collect()
 }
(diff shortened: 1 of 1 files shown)


crmne (changes requested):
Long right-to-left words now get an ellipsis, but the new character loop still lets shaped text overflow. With Fastpotify’s installed fonts at 18 px, `لالالالالالا` in a 25 px column produces `لالالا…`, which renders at 43.53 px. Subtracting isolated character widths from a shaped word’s width is not valid. Please measure the remaining shaped text as it is shortened and add a regression using the actual layout path.

Please also include before/after demo captures of a long Arabic or Hebrew title at normal and narrow widths in both themes; character-count tests cannot verify this rendering change.

kevin9327:
Thanks, you were right. Isolated widths can't describe what is left of a joining word, and the Hebrew case only looked fine because Hebrew letters don't join.

**What changed** (ea49024, 8e1b303)

- Each candidate row is measured the way `layout` draws it: the row plus the mark, through `display_line`, laid out whole with `layout_no_wrap`. The loop that gives up whole words uses the same measurement.
- When a single word has to be cut, the cut is found by bisection over boundaries that never separate a mark (NSM, BN) from its letter or fall right after a ZWJ. A piece is kept only once it has been measured inside the column.
- `a_shaped_word_longer_than_the_column_is_drawn_inside_it` goes through the real path: `theme::install`, then `layout`, then the drawn row's width against the column, for an Arabic and a Hebrew word on one and two rows, plus a check that the word stays whole and unmarked when it has room. The columns are shares of the word's own shaped width. My first push used fixed 25, 40 and 60 px columns, and with the macOS and Windows runner fonts the Arabic word fit in one of them, so 8e1b303 takes the columns from the word instead.

**Verification**

On the unfixed head, with the fonts the app installs on Windows 11 (Noto Sans Arabic) at 18 px, the new test fails with:

```
لالالالالالا in a 25.1375 px column is drawn as …لالال, 49 px wide
```

With the change, all 11 `bidi` tests pass, and `cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` are clean with `--no-default-features --features demo` (milkdrop needs vcpkg glew, which this machine doesn't have). CI is green on 8e1b303: quality, docs, the Nix package, and the tests on Ubuntu, macOS and Windows.

**Captures**

Demo mode, `--demo-page playlist:pl1`, window 1240x800 and 760x800 at 150% scaling, dark and `--demo-show light`. For the captures only, I renamed two demo tracks locally to a long lam-alef word (row 1, also in the player bar) and a long Hebrew word (row 2); that rename is not part of this PR. On this machine `--demo-shot` wrote a 1860x1350 frame whatever `--demo-size` asked for, while the window itself had the requested size, so these are the live window captured with `PrintWindow`.

Look at row 1 and the player bar. Before, the last lam-alef runs past the title cell and is sliced by its clip. After, the row ends on a whole glyph inside the cell. Row 2 is the same in both, since Hebrew letters don't join.

| | Before | After |
|---|---|---|
| 1240x800, dark |  |  |
| 1240x800, light |  |  |
| 760x800, dark |  |  |
| 760x800, light |  |  |

Row 1 and the player bar at 760x800, zoomed 2x:

| Dark | Light |
|---|---|
|  |  |

crmne (approved):
Long right-to-left titles now end with a whole glyph and an ellipsis inside the text area. I checked the normal/narrow light/dark evidence and the shaped-layout regression.

Merged as 91d19e4be058bd303c98e25522e7f2a5a45bf69e. All contribution checks, native Mac/Windows suites, and updated PR CI including Nix passed. This is on main and has not been released. Thanks!

PR #341, opened 2026-09-06
Read an album's year the way format_date reads the same field

kevin9327:
## Why

`release_date` has two readers in the crate, and only one of them can survive what the other is
written to survive.

```rust
// util::format_date
let date = iso.get(..10).unwrap_or(iso);

// Album::year
.map(|date| &date[..date.len().min(4)])
```

`min` clamps the length and says nothing about character boundaries, so `year()` panics on any date
whose fourth byte falls inside a character:

```
thread panicked at src/api/models.rs:210:30:
byte index 4 is not a char boundary; it is inside '년' (bytes 3..6 of string)
```

`format_date` reads the same field, from the same responses, and answers the same input by handing
the string back. That `get` is not there by accident — it is what says this field is not trusted to
be ASCII — and `year()` is the one place that trusts it.

I should be straight about reachability: Spotify documents `release_date` as ISO-8601, and I could
not get the live API to return anything else, so I am not claiming a user can reach this today.
`year()` is called from the artist page, the collection grid and search results, so if it ever is
reached it is a crash rather than a bad label. This is a one-line change that makes the two readers
of one field agree, and it costs nothing if the field is always well formed.

## What changed

`year()` reads with `get`, the way its neighbour does. Every documented shape is unaffected:

| `release_date` | before | after |
| --- | --- | --- |
| `2024-03-15` | `2024` | `2024` |
| `2024-03` | `2024` | `2024` |
| `2024` | `2024` | `2024` |
| `20` | `20` | `20` |
| `작년` | panic | `작년` |

The last row is the only behaviour that moves, and it moves from a panic to the same answer
`format_date` already gives: the string as it came.

## Tests

`an_albums_year_reads_a_date_the_way_format_date_does` covers all three documented precisions, a
date shorter than a year (which is what `min` was there for), the boundary case, and an album with
no date at all. It also asserts `format_date` on the same input, so the two readers are pinned as
agreeing rather than merely both being called.

Against `main` it panics at `models.rs:210`.

## How I tested

`cargo test --locked --all-targets` (348 passed, up from 347 by the new test) and

### DIFF
--- a/src/api/models.rs
+++ b/src/api/models.rs
@@ -201,9 +201,13 @@ pub struct Copyright {
 
 impl Album {
     pub fn year(&self) -> Option<&str> {
+        // `get` rather than a slice, the way `util::format_date` reads the
+        // same field: `min` clamps the length and says nothing about
+        // character boundaries, so a date whose fourth byte is inside a
+        // character is a panic rather than a date that does not parse.
         self.release_date
             .as_deref()
-            .map(|date| &date[..date.len().min(4)])
+            .map(|date| date.get(..4).unwrap_or(date))
     }
 
     pub fn kind_label(&self) -> &'static str {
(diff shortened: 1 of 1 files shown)


crmne (changes requested):
Integrated in [7b0a6ff](https://github.com/crmne/fastpotify/commit/7b0a6ff) with LF line endings and your regression test. Malformed album dates no longer crash the app. Thanks, @kevin9327. This will be in the release after 0.7.1.

Issue #327, opened 2026-09-05
Baseline misalignment between CJK and Latin characters (e.g. Japanese + English text) on macOS


future10se:
### Fastpotify version and how you installed it

0.6.0 from DMG

### Operating system

macOS

### System details

macOS Tahoe 26.6.2 / macOS Golden Gate 27 Beta 8

### Spotify plan

Premium

### What did you do, and what happened instead?

1. Have songs with Japanese/Chinese titles, album, or artists together with English text (or any language with latin characters)
2. View any list (e.g. Library, Liked Songs, Playlists, etc.)
3. Note that the CJK characters don't sit on the same baseline as latin characters (they are offset upwards)

I noticed this only occurs on my **macOS** machines (latest patches of 26 stable and 27 beta).

It does **not** happen on my Windows 11 PC.

See the screenshot below:

<img width="345" height="306" alt="Image" src="https://github.com/user-attachments/assets/7f66f3f8-35d3-402f-91e6-25e793754420" />

The most obvious example above is track 87.

Also note, even without mixed text, it seems CJK characters by themselves are offset. (e.g. see track 86, there is less space at the top of the cell vs. the bottom)

This happens anywhere text is rendered as far as I can tell, including artist profile pages, lyrics, etc.

<img width="173" height="150" alt="Image" src="https://github.com/user-attachments/assets/b023c813-2951-41b8-9f59-19147c3d6db0" /> <img width="267" height="150" alt="Image" src="https://github.com/user-attachments/assets/63816b00-19d3-423e-bb81-b3ba66ecb435" />

<img width="529" height="125" alt="Image" src="https://github.com/user-attachments/assets/2a2fbb1c-03d8-4e4b-9464-f38b00f06aca" />

<img width="288" height="371" alt="Image" src="https://github.com/user-attachments/assets/13e9a0a4-57ad-45e6-b78f-e1f9780e176e" />

### fastpotify.log

```text
<log file was empty, just launch the app and scroll through the library. you don't need to start playback or anything to see the issue.>
```

crmne:
#334 has merged a macOS font-selection fix that may help here. Could you retest the mixed Japanese/Chinese and English titles with a build from main, or the next release, and confirm whether their baselines now align? I’m keeping this open until that specific problem is verified.

future10se:
I built the latest commit (03c3e69) and am unfortunately still running into this issue.

[...]

PR #345, opened 2026-09-06
fix(theme): align fallback font baselines with Inter

kveld9:
### Problem

When titles contain mixed Latin and CJK text (or pure CJK text), characters from system fallback fonts sit higher than Latin text instead of resting on the same baseline (#327). This occurs noticeably with Hiragino Sans on macOS when Japanese is among the user's preferred languages.

### Root Cause

In `epaint` (`text_layout.rs`), glyphs are positioned vertically using:

```rust
glyph.pos.y = glyph.font_face_ascent + 0.5 * (primary.row_height - fallback.row_height);
```

`epaint` centers the difference in line height between the primary font (Inter) and the fallback face rather than aligning their baselines.

- **Inter**: Ascent is 0.969 em, line height is 1.210 em. The baseline-to-center offset is `0.969 - 0.5 * 1.210 = 0.364 em`.
- **PingFang SC**: Ascent is ~1.06 em, line height is ~1.36 em. Its baseline-to-center offset is `1.06 - 0.5 * 1.36 = 0.380 em` (a 0.016 em delta, appearing visually aligned).
- **Hiragino Sans**: Apple defines an ascent of 0.88 em and a total height of 1.50 em via a 0.50 em `lineGap`. Its baseline-to-center offset is `0.88 - 0.5 * 1.50 = 0.130 em`.

The resulting delta (`0.364 - 0.130 = +0.234 em`) shifts Hiragino glyphs upward by ~3.3 pt at a standard 14 pt font size.

### Solution

In `src/theme.rs`, read each fallback font's vertical metrics with `skrifa` and set `data.tweak.y_offset_factor`:

```rust
(inter_ascent_ratio - 0.5 * inter_height_ratio) - (fallback_ascent_ratio - 0.5 * fallback_height_ratio)
```

This offsets the glyph downward to neutralize `epaint`'s centering calculation, aligning the baselines across all mixed scripts and font sizes on all platforms.

### Verification

- `cargo fmt --all --check`
- `cargo test --locked --all-targets --all-features` (359 library tests and 5 binary tests pass, including new tests for Inter baseline identity and mixed CJK text layout)

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -313,6 +313,10 @@ fn install_fonts(ctx: &egui::Context) {
         // epaint rebuilds the glyph atlas.
         let mut data = FontData::from_static(&font.bytes);
         data.index = font.index;
+        let offset = fallback_baseline_y_offset(&font.bytes, font.index);
+        if offset.abs() > 0.001 {
+            data.tweak.y_offset_factor = offset;
+        }
         fonts.font_data.insert(font.name.clone(), Arc::new(data));
         for family in fonts.families.values_mut() {
             family.push(font.name.clone());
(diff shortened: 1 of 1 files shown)


crmne:
Visual change: system fallback text now aligns with Inter, including mixed Japanese/Chinese and Latin titles. Merged as 46dc449f688166f193af38b25df8e7854900ad83. Thank you!

I reproduced the Hiragino offset on the Mac, added a regression for the painted glyphs, and checked narrow/normal windows in both themes. All required checks, native Mac/Windows tests and hosted CI passed. This is not released yet.

Issue #79, opened 2026-08-29
Install updates from the app without replacing package-managed installations


charliealgernon:
### Before requesting

- [x] I searched open and closed issues for this request.
- [x] I read the project boundaries in CONTRIBUTING.md.

### What problem should Fastpotify solve?

When I install Fastpotify (Homebrew cask, AUR, DMG, or a standalone binary), the app never tells me that a newer release exists. I only find out by checking GitHub Releases myself, or by noticing a missing fix after I already hit a bug.

That is easy to miss. Releases are moving quickly (`v0.1.x` through `v0.3.0` and `v0.4.0-rc1` in a few days), so people stay on old builds without knowing it.

There is no in-app version check, no update notification, and no way for a standalone/DMG install to apply a new release without leaving the app.

### What would a good solution look like?

Two parts, both explicit:

**1. Check for a new version**

- On launch (and optionally on a long interval, e.g. once a day), compare the running version to the latest *stable* GitHub Release.
- If a newer version exists, show a clear in-app notice: current version, new version, and a short changelog or release notes link.
- Do not use pre-releases (`-rc`, `-beta`) unless the user is already on a pre-release or opts in.
- Fail quietly if GitHub is unreachable. Never block startup or playback on the check.
- Make the check optional in Settings (on by default is fine). No extra Fastpotify service and no telemetry — just the public GitHub Releases API.

**2. Apply the update**

Behaviour should match how the user installed the app:

- **Standalone binary / DMG / Windows installer:** offer **Update now**. Download the matching asset for this OS/arch, verify it (checksum / signature if you publish one), replace the current install, and restart. If a silent auto-download is added later, still ask before replacing the running app.
- **Homebrew, AUR, Nix, `cargo install`:** do **not** overwrite the package-managed binary. Tell the user a newer version exists and show the exact command for their install path (`brew upgrade --cask fastpotify`, `yay -Syu fastpotify`, etc.), plus a link to the release.
- If the install method cannot be detected, fall back to “a new version is available” + the release page. Never guess and overwrite a package-manager install.

Also useful: a **Check for updates** action in Settings (or the Help/About area) so the user can trigger this without restarting.

### Why does this fit Fastpotify?

This is ordinary desktop-client behaviour, not a new product direction. Fastpotify already ships native builds for Linux, macOS, and Windows; those builds should be able to stay current without a browser or a Fastpotify-operated update service.

[...]

PR #379, opened 2026-09-10
Add self-update for Linux hand-installs

haastt97:
## Summary

Users can now update Fastpotify from within the app. On Linux, the download is verified against the release's SHA-256 checksums, extracted over the running binary, and re-executed — all without leaving the app.

- "Update to X.Y.Z" pill in the top bar + "release notes" link
- Settings → About → "When an update is found" dropdown:
  - Manual: only notify when a newer release exists
  - On Click: download when the user clicks the update pill
  - On Launch: auto-update on startup before the window appears
- Non-Linux platforms open the release page when the user clicks

## Notes

- Changes were tested on v0.7.0 and v0.7.1
- Longcat 2.0 wrote the code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,7 +271,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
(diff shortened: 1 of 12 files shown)


crmne:
This changes startup to install updates automatically by default and replaces the executable without checking package-manager ownership. That does not meet #79’s requirements. I’m choosing #383 as the updater implementation to review and closing this competing approach. Thank you for working on it.

PR #383, opened 2026-09-10
Add in-app updates for Windows, macOS, and Linux

coah80:
Fastpotify currently sends users to the release page when they click the update pill. This adds downloading and installing updates inside the app on Windows, macOS, and Linux. Restarting always stays under the user's control.

**Tested on Windows, macOS, and Linux**, including downloading a real GitHub release, verifying the installed files, and opening the updated app.

### What changes for users

- Clicking **Update to …** opens a small update window styled with the existing app palette, typography, and buttons. Update checks never open it automatically.
- **Download update** shows download progress and verifies the file before offering **Restart to update**. Closing the window keeps the download running.
- **Download updates automatically** is an optional setting, off by default. It downloads in the background and waits for the user to restart.
- **Release notes** opens the GitHub release page for the offered version.
- Failed downloads offer **Retry download** and leave the running installation intact. The restart screen explains that local music playback will stop.

### Installing the update

The downloader selects the release asset for the operating system, CPU architecture, and installation type. It checks the release version, published file size, and SHA-256 checksum before staging the update beside the installation.

| Installation | Update behavior |
| --- | --- |
| Windows portable | Replace the executable after the app exits, then restart it |
| Windows installer | Run the downloaded installer against the existing installation, then restart |
| Linux portable | Replace the executable after exit, then restart |
| macOS app bundle | Install the complete bundle from the universal DMG after checking its identity, version, and signature. Developer ID installations also require the same signing team and macOS approval |
| Package-managed or unrecognized | Show package-manager guidance or the release download page |

A helper keeps a backup and waits for the updated app to confirm that its first window frame has rendered. If startup fails or times out, it restores the previous executable or Mac bundle and reopens it with an error message. Settings and sign-in files remain in place.

Release packaging adds installation markers so portable downloads and Windows installer builds can be identified. Older releases need one manual installation of an update-enabled build before using this flow. No new dependencies are added.

### Live-release test

Downloading and installing the actual 0.7.1 release on Windows

The Windows test starts from an isolated development build labelled **0.7.0** and installs the **actual GitHub 0.7.1 release**. Published 0.7.1 predates the new startup-confirmation protocol, so that particular test uses an external native-window and executable-hash check. That compatibility override is confined to the test copy and is **not included in this PR**. The normal startup confirmation and rollback paths were tested separately with updater-enabled builds.

### Validation

- Real GitHub 0.7.1 download, replacement, matching installed hashes, and visible restarted windows on Windows x64 portable, an Ubuntu 24.04 VM running locally on the Windows PC, and an ARM64 MacBook. The Mac installation also matched every file in the published bundle.
- Updater-enabled build tests covered the Windows installer, damaged downloads and retry, failed-start rollback, and background downloads without opening the popup or restarting automatically.
- Debian 13, Fedora 44, and Arch userspace tests ran inside the local VM under Xvfb, including package-manager ownership checks. These share the VM's kernel.
- Windows formatting, Clippy, and all-target tests passed with `--no-default-features --features demo`. Linux and Mac default/all-feature checks, doctests, Rust documentation, Node assessment tests, and the Jekyll build passed.

Not exercised locally: Windows/Linux ARM execution, Intel Mac execution, Wayland, Windows MilkDrop builds, or a notarized Developer ID upgrade. The public-release run used Windows portable, not the public Windows installer.

Related: #379 proposes Linux hand-install updates. This is a separate implementation covering Windows and macOS too, with explicit restart confirmation and startup rollback.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -91,6 +91,7 @@ jobs:
             cp "target/${{ matrix.target }}/release/fastpotify" "dist/$name/"
           fi
           cp README.md LICENSE "dist/$name/"
+          cp packaging/fastpotify-portable.txt "dist/$name/"
           if [ "${{ runner.os }}" = "Linux" ]; then
             # Desktop integration for package builds and hand-installs.
             cp -r packaging "dist/$name/"
--- a/README.md
+++ b/README.md
@@ -317,6 +317,27 @@ Playback settings apply when you press **Apply and restart playback**.
 You can also check for a new release from Settings. On macOS, the same command
 is in the application menu.
 
+On Windows and Linux, update-enabled portable downloads can download a release
+in the app, verify its published SHA-256 checksum, and restart to install it.
+Windows installer builds use their installer for the replacement. Settings can
+enable automatic background downloads; restarting always requires a click.
+The update popup opens only when you click the green update pill. Update checks
+and automatic downloads leave it closed, and closing it keeps downloads running.
+A failed startup restores the previous installation. An interrupted or damaged
+download leaves the running app alone. Updates keep your settings and sign-in
+files. On macOS, a writable Fastpotify.app downloaded from the release page can
+update its whole app bundle from the universal DMG. Move the app out of the disk
+image before updating. The updater verifies the app signature and version;
+Developer ID builds also require the same signing team and macOS approval.
+Keep the app in Applications; macOS can require folder access when it is run
+from Documents.
+
+Package-managed installations continue to update through their package manager,
+including Homebrew, Flatpak, apt, dnf, pacman, Nix, and Cargo. Unrecognized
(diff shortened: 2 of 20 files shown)


coah80:
@codex review

crmne (changes requested):
This adds the update window, optional background downloads and an explicit restart action. Integrated on main in 5841823d4830ee70ff6ce31db006c7b612d8ea91, with your contribution credited.

I fixed interrupted backups and macOS mountpoint retries, and integrated the setting with Settings search. Full local checks and Linux, macOS and Windows CI pass. I also verified native Linux installation and failed-start rollback, and reviewed narrow/wide popup captures in both themes. Your reported Windows and macOS runtime tests remain the native evidence for those platforms.

Not released yet. Thanks for building this.

PR #384, opened 2026-09-10
Keep playlist and album scrollbars at their full length

alexng353:
## Why

- Album and playlist scrollbars grow as tracks load, so their size and position do not represent the finite collection. Users cannot drag directly to a distant track.
- Matches behaviour with official client

## What changed

- Reserve Spotify's total track count immediately and fetch the visible 50-row window when scrolling. Reuse cached windows and keep loaded rows visible across page boundaries.
- Preserve server positions for unavailable tracks, keep playback and playlist edits aligned, and show Retry for failed windows. Update pagination and cache documentation.

## Verification

- Linux: `cargo fmt --check`, `cargo test --all-targets`, `cargo test --all-targets --all-features`, and `cargo clippy --all-targets` with and without `--all-features`, both with `-- -D warnings`, passed on this upstream-based branch.
- Regression tests cover direct jumps, overlapping windows, cache races, filtering, edits, unavailable tracks and retries. No signed-in Spotify, macOS or Windows testing.
- Doctests, rustdoc, Node issue-assessment tests and Jekyll passed before changing the base to upstream/main.
- Native demo below: 50 loaded tracks out of 1000. Captured against the original feature parent; additional themes/sizes and capture provenance are in `docs/assets/finite-scroll/`.

| Before | After |
| --- | --- |
| Before | After |

Co-Authored-By: GPT-6 Astra <noreply@openai.com>


crmne (commented):
Playlist and album scrollbars now represent the full collection in f70231a4ef0b9ab06148d2fc08ff580910f24ce9, with distant loading, placeholders and Retry. The approved design, local checks and hosted Linux, macOS, Windows and Nix checks pass. Signed-in Spotify verification is not claimed. This has not been released yet. Thanks for the contribution.

PR #387, opened 2026-09-10
Combine the personal and shared apps for search

kglennon:
## Summary

- Today, search always goes through the shared, often congested Web API quota alone, for the complete, playlist-inclusive result set.
- There's an opportunity to combine both apps for a single search instead: the personal app can already serve almost everything a search needs, fast, and the shared app is only truly required for the slice of results Spotify itself owns (its editorial/curated playlists, like Today's Top Hits).
- Verified live against a real account: Spotify's search endpoint actually just **omits** Spotify-owned/editorial playlists from a personal app's results rather than rejecting the request. Everything else (tracks, artists, albums, user/community playlists) comes back fine, and fast.
- How much faster, measured live against a real account during ordinary shared-app congestion: individual shared-app requests took **8 to 29 seconds** (`8699ms`, `10518ms`, `21527ms`, `22433ms`, `28772ms`, and similar, across many samples), while the same kind of request through a personal app consistently completed in **under 1 second** (mostly 100-900ms). Not a controlled benchmark, just what live testing this session actually saw, but the gap was consistent and large across many searches.
- No new setting. Search now routes through `Operation::Catalog`, the same category as `Playback`, `UserData`, and other lookups that already use the personal app when one is ready and fall back to the shared app on their own otherwise.

## How it actually behaves

- **No personal app configured:** unchanged from before this PR. One request, through the shared app, covering everything including Spotify-owned playlists.
- **Personal app configured and ready:** two requests fire concurrently for every search:
  1. The main lookup (tracks, artists, albums, community/user playlists, shows, episodes) through the personal app. Fast, and it's what the listener sees first.
  2. A second, playlist-only lookup through the shared app, specifically to find Spotify-owned/editorial playlists the personal app cannot see. Its results are merged into the already-displayed results once it returns, bounded by the shared app's own `MAX_RETRY_AFTER` (30s), the same cap its rate-limit cooldowns already use. This second request never blocks the first, and if it fails or times out, the personal-app results already shown are left as they are.
  A newer search cancels a still-in-flight (2) from an earlier keystroke, so a listener correcting a typo doesn't stack up several of these against the shared app's quota.

  Spotify's own playlists effectively stream in from the shared app: they appear in the Playlists section shortly after everything else, once that second lookup resolves. There's no dedicated loading indicator for this in-progress merge right now: the top-level spinner already stops once the personal app's results are shown, and the Playlists section just stays absent until it either fills in or the lookup gives up. Worth a follow-up if that gap in the loading state bothers anyone; not addressed here.

## Test plan

- [x] `cargo test --locked --all-targets` (461 to 468 passing, 6 new: `operation_for` routing, `is_spotify_owned` classification x3, editorial-playlist merge behavior x2)
- [x] `cargo clippy --locked --all-targets -- -D warnings` clean
- [x] `cargo fmt --all --check` clean
- [x] Manually verified live against a real Spotify account: "pulp" returned real playlists from non-Spotify owners via the personal quota; "Today's Top Hits" correctly merged in the real "Today's Top Hits," "Hot Hits UK & Ireland," and other `By Spotify` playlists once the merge fetch completed
- [x] Full CI on GitHub Actions: quality, docs, Linux/Windows/macOS tests, and the Nix package build all passed on this final version (https://github.com/kglennon/fastpotify/actions/runs/34506317137)

### DIFF
--- a/src/api/client.rs
+++ b/src/api/client.rs
@@ -21,7 +21,7 @@ use super::models::*;
 const BASE_URL: &str = "https://api.spotify.com/v1";
 const MAX_IN_FLIGHT: usize = 6;
 const RATE_LIMIT_RETRIES: u32 = 3;
-const MAX_RETRY_AFTER: Duration = Duration::from_secs(30);
+pub(crate) const MAX_RETRY_AFTER: Duration = Duration::from_secs(30);
 
 #[derive(Clone, Debug, Error)]
 pub enum ApiError {
--- a/src/app.rs
+++ b/src/app.rs
@@ -1303,6 +1303,11 @@ impl App {
                 }
                 Event::Local(state) => self.handle_local(*state),
                 Event::Api(response) => self.handle_api(*response),
+                Event::EditorialPlaylists {
+                    query,
+                    serial,
+                    playlists,
+                } => self.merge_editorial_playlists(query, serial, playlists),
                 Event::Accent { url, color } => {
                     self.accent_pending.remove(&url);
                     let tint = self.palette.tint_from_art(color);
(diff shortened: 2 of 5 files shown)


crmne:
Thanks for working on this. I merged the search implementation in #388, including cancellation based on your approach and contributor credit. It keeps full shared-app playlist coverage and handles either request finishing or failing first. Closing this competing implementation now that the combined solution is on main.

PR #388, opened 2026-09-10
Search the catalogue through the personal app

joshkat:
When a personal app is connected, catalogue search now uses its quota while a separate shared-app request finds playlists, including Spotify-owned playlists. Without a personal app, search keeps the existing six-type shared request.

Each successful part appears as soon as it arrives and remains usable if the other fails. Loading and failure messages reflect the unfinished or failed part. Starting or clearing a query cancels the previous requests; late results cannot mix songs and playlists from different queries. The request type keeps its scope if personal access changes while it is waiting. Personal catalogue pages contain ten results per type, compared with twenty through shared access.

The README and connection guide describe the routing and limits. Regression coverage includes both arrival orders, either or both halves failing, cross-query results, clearing, sign-out, cancellation while shared access is unavailable, and personal-grant readiness changes. A demo test checks that an unfinished or failed search is not reported as having no results.

All CONTRIBUTING checks pass locally on Linux. All PR CI jobs pass, including Linux, macOS, Windows and Nix. The combined changes also pass the full local checks against current main. Cancellation follows the approach proposed by @kglennon in #387, with contributor credit included.


crmne (approved):
Catalogue results and playlists now appear independently, with loading and errors shown for each request. This scope is approved. The failure handling, query identity, cancellation and routing regressions pass, as do all contribution checks on the combined main tree and all PR CI jobs. Contributor credit for the cancellation approach from #387 is retained.

PR #389, opened 2026-09-10
Update README with Gentoo installation instructions

NikoMalik:
Added installation instructions for Gentoo users

## Why

Installation on gentoo via package

## What changed

Only change in readme


crmne (changes requested):
Integrated in 5d571a73b2dbb36667cf976f390563c75cf19ae9. I corrected the install section and documented the community overlay as snapshot 67b8dfb without MilkDrop. All PR checks passed, including Nix. Thanks for adding Gentoo instructions.

PR #397, opened 2026-09-12
Add a refresh control on playlist pages

AQS-DS:
## Why

I was using Fastpotify and liked how focused and fast it feels. After adding a
few songs to a playlist outside the app (or from another client), those tracks
never showed up no matter what I tried inside Fastpotify. Closing and reopening
the app was the only reliable way to see the update.

Keeping playlists from constantly re-fetching makes sense for performance, and
that priority is part of why this project works well. Still, needing a full
restart just to refresh one playlist is awkward. A small Refresh control on the
playlist page felt like the right balance: on demand, no background polling, and
no change to the rest of the app.

## What changed

- `Actions` in `src/ui/collection.rs` accepts an optional `reload: Option<(Page, bool)>`.
- Playlist pages pass that page and its loading flag; album and Liked Songs leave it `None`.
- A Refresh control sits next to the Filter field.
- While reloading, the control shows a spinner, is not clickable, and uses the Refreshing… tooltip.
- Uses the existing `Action::Reload(Page)` path.

## Verification

- Platform: Windows
- `cargo check --locked --no-default-features`
- `cargo run --locked --no-default-features`
- Opened a playlist, clicked Refresh, confirmed the spinner and updated tracks.

### Screenshots
**Before**
F brefore

**After**
F after

**Refreshing**
F while

## Relation to #271

This is complementary to #271 rather than dependent on it. #271 changes how eligible playlist reads are served; this PR provides an explicit way to trigger a new read for a playlist that is already loaded in memory.

If #271 lands, manual refreshes can benefit from its session-backed read path automatically. The control remains useful independently of it, since it only triggers the existing `Action::Reload(Page)` path and adds no background polling.

Docs are unchecked on purpose until the Refresh control placement is confirmed in review.

### DIFF
--- a/src/ui/collection.rs
+++ b/src/ui/collection.rs
@@ -270,11 +270,10 @@ pub fn actions_row(
                             )
                         });
                         if ui.is_rect_visible(rect) {
-                            let mut child = ui.new_child(
-                                egui::UiBuilder::new().max_rect(rect).layout(
+                            let mut child =
+                                ui.new_child(egui::UiBuilder::new().max_rect(rect).layout(
                                     Layout::centered_and_justified(egui::Direction::LeftToRight),
-                                ),
-                            );
+                                ));
                             theme::spinner(&mut child, 22.0, palette.secondary);
                         }
                         response.on_hover_text("Refreshing…");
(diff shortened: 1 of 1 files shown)


crmne:
Refresh is now in the playlist’s **…** menu in 6f1c4eacf3b165337ff1919555cb05ee76e48f4b. It uses the existing reload path and is disabled while loading. All local checks, the visual review and hosted Linux, macOS, Windows and Nix checks pass. This has not been released yet. Thanks for the contribution.

PR #399, opened 2026-09-13
Package the Linux builds as deb and rpm

aljazceru:
## Why

Linux users outside Arch and Flatpak currently get a tarball they must
unpack by hand, with no desktop entry and no dependency resolution. Ubuntu
and Fedora users can install native packages, so the release should carry
them.

## What changed

The release workflow packages the Linux binary it already builds as a
`.deb` and an `.rpm`, one per architecture (`x86_64`, `aarch64`), with
`cargo-deb` and `cargo-generate-rpm`. The binary is the tarball's, so
nothing compiles twice and the packages carry the build's glibc baseline:
Ubuntu 24.04 or Debian 13 and newer for the deb, Fedora 41 and newer for
the rpm. Checksums cover them automatically, and the README and download
page document the new install commands.

`Depends` and `Requires` live in `Cargo.toml` metadata. They list both
what the binary links (libstdc++, ALSA, PulseAudio) and the Wayland, X11,
and xkbcommon stacks the GUI loads at run time, which a package manager
cannot discover. Fedora splits its GL libraries across `libglvnd-glx` and
`libglvnd-egl`, and `libxkbcommon-x11` is a separate package there.

## How I tested

- Built both packages with the exact workflow commands from the branch.
- Installed them in containers of each distro's minimum version
  (`fedora:41`, `ubuntu:24.04`) and current `fedora:latest`: dependencies
  resolve, `fastpotify --version` runs, and every library the GUI loads
  at run time is present afterwards.
- `readelf -d` confirms the linked sonames; the highest glibc symbol the
  binary needs is `GLIBC_2.39`, matching the documented minimums.
- The full check battery from CONTRIBUTING on this branch: fmt, both
  clippy runs, both test runs (516 and 525 passing), doc tests, rustdoc
  with `-D warnings`, the node assessment tests, and the Jekyll docs
  build.


aljazceru:
Addressed in 3631b8f: the x86_64 Linux leg is pinned to ubuntu-24.04 so the package baseline matches what the README and Cargo.toml promise, cargo-deb and cargo-generate-rpm are pinned to the versions the packages were built and container-tested with (3.8.0 and 0.21.0), and the README commands now mention the aarch64 artifact as the download page does.

62b9582 then rewrites rpm requires as a TOML sub-table: the multi-line inline table parsed for cargo but not for Nix's stricter fromTOML, which failed the Nix package job. Both parsers and cargo-generate-rpm accept the new form; the generated requires are unchanged.

crmne:
Thank you, @aljazceru. I adapted your runtime dependency findings and runner baseline into the new native-packages pipeline in aa9b429905f721039ba5ab556a0e54d1f9abac17, with co-author credit. Installation, runtime-library loading and removal pass on Ubuntu, Debian and Fedora for both architectures, and all application CI checks pass.

Closing this PR because its work is now covered on main. The packages will ship with the next release; they are not released yet.

PR #415, opened 2026-09-13
Look for Shift shortcuts before the plain ones they extend

kevin9327:
## Why

On Windows and Linux, Ctrl+Shift+Q is listed as "Show the queue", but pressing it quits the app. Ctrl+Shift+B, listed as "Go to the playing album", toggles the sidebar instead, on every platform (Cmd+Shift+B on macOS).

`consume_key` matches modifiers with `Modifiers::matches_logically`, which ignores an extra Shift or Alt. The egui docs say to check the more specific shortcut first for exactly this reason. In `keys::handle`, the plain Ctrl+Q (`Quit`) and Ctrl+B (`ToggleSidebar`) are consumed before the Shift variants are checked, so they take the key first and the Shift shortcuts can never fire.

## What changed

In `src/ui/keys.rs`, the three Shift shortcuts that extend a plain one (Ctrl+Shift+A, Ctrl+Shift+B, and Ctrl+Shift+Q, or Cmd+U on macOS) are now looked for before the plain Command shortcuts, with a comment saying why the order matters. The set of shortcuts and what each one does are unchanged, and `SHORTCUTS` already describes the intended behaviour, so there is no documentation change. The Winamp window calls the same `keys::handle`, so it is fixed too.

## Verification

`a_shift_shortcut_is_not_taken_by_the_plain_one_it_extends` sends key events through `keys::handle` on a demo-populated `App`, using the modifiers the platform reports for its command key. It checks that Ctrl+Shift+Q shows the queue (not on macOS, where that key is Log Out), Ctrl+Q still quits, Ctrl+Shift+B opens the playing album, and Ctrl+B still toggles the sidebar.

```
assertion `left == right` failed: the queue shortcut shows the queue
  left: "[Quit]"
 right: "[ToggleQueuePanel]"
```

and, with the Q checks skipped:

```
assertion `left == right` failed: the album shortcut goes to the album
  left: "[ToggleSidebar]"
 right: "[Open(Album(\"alb0\"))]"
```

- `cargo fmt --all --check`

Tested on Windows only. The macOS branch of the test compiles in the same code but was not run here. No screenshots, since nothing drawn changes.


crmne:
Integrated as 34da86233cb5207dd5955c5a6e584650f06f3d6a, preserving your contributor credit. All required checks passed, including Nix, plus native Mac and Windows tests. This fix is on main and is not released yet.

PR #416, opened 2026-09-13
Leave the arrow keys to a focused text field

kevin9327:
## Why

Typing in the search field and pressing Ctrl+Left to go back a word skips to the previous song. Ctrl+Right skips to the next one, Ctrl+Up and Ctrl+Down change the volume, and Alt+Left and Alt+Right navigate Back and Forward. The caret does not move. On macOS the same happens with Cmd+Left and Cmd+Right (start and end of the line) and Option+Left and Option+Right (by word).

`keys::handle` runs at the start of the frame, before any page draws its text fields, and consumes those keys for the app's shortcuts. egui's `TextEdit` uses the same keys to move its caret (Ctrl or Alt by a word, Cmd to the end of the line, Command with Up or Down to either end of the text), so the field never sees them. The plain shortcuts that clash with typing (Space, M, Shift+arrows for seeking, and so on) are already skipped while something has focus. The Command and Alt arrow shortcuts are not.

## What changed

In `src/ui/keys.rs`, the six arrow shortcuts (Back, Forward, Previous, Next, and volume up and down) are skipped while a text field has focus, using egui's `Context::text_edit_focused`. That is narrower than the existing `typing` check on purpose: a focused song row is not a text field, so Ctrl+Right still changes songs from the keyboard-navigated song list. Every other shortcut is unchanged, and the shortcuts dialog already describes these keys as player shortcuts, so there is no documentation change.

## Verification

`a_focused_text_field_keeps_the_arrow_keys_it_edits_with` draws a frame the way `ui::show` does, `keys::handle` first and then a `TextEdit`, on a demo-populated `App`. With the field focused and the caret at the end of `find this song`, it presses Ctrl+Left (Cmd+Left on macOS), Alt+Left and Ctrl+Up. It asserts that no action is emitted, that the caret moved where egui moves it, and that the text is unchanged. It then takes focus away and checks that Ctrl+Right is Next again and Alt+Left is Back again.

```
assertion `left == right` failed: the previous-song shortcut took ArrowLeft from the focused field
  left: "[Previous]"
 right: "[]"
```

- `cargo fmt --all --check`

Tested on Windows only. The macOS expectations in the test (Cmd+Left to the start of the line) follow egui's own cursor rules but were not run here. No screenshots, since nothing drawn changes.

This touches the same function as #415 (a different change: the Shift shortcut order). Whichever lands second may need a small conflict resolved, and I will take care of it.


crmne:
Integrated in 081bd449f887bd99b21754871ce638ac934b7b99. Text fields keep caret-navigation arrows, and playback shortcuts still work outside them.

PR #421, opened 2026-09-13
Keep a short song played twice as two plays in Recent

kevin9327:
## Why

Play a short song twice in a row, for example a forty-second interlude, on this computer or on another device, and the Recent tab shows it once. The second play is dropped.

`history::merged` combines the plays made here with Spotify's recently played list. To avoid showing one play twice when both sources report it, it treats any two plays of the same song that start within 60 seconds of each other as one. But it applies that window to every pair, including two plays from the same source, and a song shorter than a minute played twice starts its second play less than 60 seconds after the first. `App::absorb_recents` already documents the intended rule for Spotify's list: "Repeated plays remain separate. Only identical track-and-time entries are deduplicated across page boundaries." The merge then undoes that. Plays made here are recorded once per listen by `note_listening`, so there are no duplicates within the local list to remove either.

## What changed

In `src/history.rs`, `merged` now uses the 60-second window only across sources: a play Spotify reports is dropped when it falls within the window of a play made here, and each local play stands for at most one report (the closest). Plays from the same source are never merged. The ordering, the handling of plays without a timestamp and the window itself are unchanged, and the doc comments on `merged` and `SAME_PLAY` now say this. The user guide already describes the tab as combining the two histories, so it needs no change.

## Verification

`a_short_song_played_twice_by_one_source_is_two_rows` calls `merged` with two plays of the same song 40 seconds apart, first as Spotify's list, then as the local list, and expects two rows each time. It then gives both sources those two plays (5 seconds apart across sources) and expects two rows, each play once. The existing `one_play_seen_twice_is_one_row` still checks that one play reported by both sources is one row, and that plays five minutes apart stay separate.

```
assertion `left == right` failed: two plays on another device
  left: 1
 right: 2
```

- `cargo fmt --all --check`

The change is plain `std` collections with no platform-specific code; it was run on Windows only. No screenshots, since the tab's layout does not change, only which rows it lists.


crmne:
Integrated in 1c1586b19b54ef6f1e785a20b4bae4e5df0bd1da. Recent keeps repeated short-song listens, including consecutive local repeats. Each repeat earns its own listening time.

PR #422, opened 2026-09-13
Play a Recent row's own song whatever its place in the tab

kevin9327:
## Why

Double-click any song in the Recent tab except the top one, and the interface does not show it as starting: the player bar waits for the engine to report the new song instead of switching at once, as it does for every other list. When a Connect device is the target, the Web API is sent an offset past the end of the list it is asked to play.

Each Recent row plays a list holding only its own song, `RowContext::Uris([song])`, but passes its place in the whole tab as the index (`src/ui/queue.rs`). `Action::PlayFromRow` then builds `PlayRequest::tracks([song]).starting_at_index(n)`, and for every row but the first, `n` is past the end of that one-song list. `play_request` looks up the song at `n` to show it as starting, finds nothing, and clears the intent. For a remote device the request goes out with `offset.position` set to `n`; I could not test how Spotify answers that from here. Local playback still starts the right song, because `local_load` loads a single song by its URI and drops the index.

The row index cannot just be 0: rows without a number, as in the Recent tab, use it for their widget and menu ids.

## What changed

In `src/app.rs`, the `RowContext::Uris` arm of `Action::PlayFromRow` keeps the index when the song at that place is the song the row names, and otherwise starts at the named song's place in the list. This is the rule Next up rows already follow in `play_queue_item` ("if the rows shifted under the pointer, the song wins over the row number"). Rows whose index already points at their song, including a repeated song in a list, are unchanged.

## Verification

`a_recent_row_plays_its_own_song` goes through the real path. It fills the Recent tab with three plays, draws `ui::queue::side_panel` with the Recent tab open on a demo-populated app whose local engine is still connecting, clicks the second row through AccessKit, and applies the actions. It asserts that the held request plays that song from position 0 and that the song is the one shown as starting. It also applies a row whose list repeats a song and checks that the row's own position is kept.

```
assertion `left == right` failed: the song's place in the list it plays from
  left: Some(1)
 right: Some(0)
```

- `cargo fmt --all --check`
- `cargo test --locked --no-default-features --features demo --all-targets` (all lib tests pass, including the new one)

Run on Windows only; the change has no platform-specific code. No screenshots, since nothing drawn changes.


crmne:
Integrated in d0f8c4de2e921221946ed11f2dc61f3d3ce62415. Choosing any Recent row plays the song it names and shows it immediately in the player bar.

PR #444, opened 2026-09-14
Queue a song picked twice as two rows

kevin9327:
## Why

A playlist can hold the same song more than once. Picking both of its rows
and choosing **Add to queue** from the right-click menu queued the song only
once, while the notification said every picked song had been added. The
queue's rules in `docs/_reference/queue.md` (rule 2) say that queuing the same
song twice plays it twice, and that only a double-click counts once.

## What changed

The picked-rows menu sends `Action::QueueMany`, which adds each song through
`queue_one`. That function also held the double-click guard: a song queued
within the last 1.5 seconds is ignored. Inside one batch, the second copy of a
song found the pending add made for its first copy a moment earlier and was
dropped as if it were the same click arriving twice.

The guard now lives in `queued_moments_ago`. A single Add to queue checks it
exactly as before. `QueueMany` decides which of its songs repeat an earlier
click before it adds any of them, so rows from the same menu click no longer
count against each other, and the same batch arriving twice still queues
nothing more.

No documentation change: rule 2 already describes the corrected behaviour.

## Verification

New test `app::tests::a_song_picked_twice_is_queued_twice` queues `b, c, b`
through `Action::QueueMany`, then sends the same batch again. On `main` it
fails:

```
assertion `left == right` failed: every picked row is queued, in order
  left: ["spotify:track:b", "spotify:track:c", "spotify:track:ctx1"]
 right: ["spotify:track:b", "spotify:track:c", "spotify:track:b", "spotify:track:ctx1"]
```

With the change it passes, and `asking_play_next_twice_queues_two_rows` still
passes, so a double-clicked Add to queue still counts once.

Run on Windows 11:

- `cargo fmt --all --check`
- `cargo clippy --locked --all-targets --no-default-features --features demo -- -D warnings`
- `cargo test --locked --no-default-features --features demo --lib` (619 passed)

The default MilkDrop feature needs GLEW, which this machine does not have, so
the default and all-features builds are left to CI. Not tested on macOS or
Linux; the change has no platform-specific code.


crmne:
Thanks, this is integrated in 8ea2b9953c10bdacd337bf990475abf4395e9b2a. Picking the same song twice in a selection now adds both rows, and repeat events no longer show an extra “added” message. All local checks and Linux, macOS, Windows and Nix CI pass. This is on main and has not been released yet.

PR #445, opened 2026-09-14
Keep older history pages behind newer plays in the sidebar order

kevin9327:
## Why

With the Library sorted by **Recently played**, the sidebar follows
`App::recent_contexts`, which is documented as the contexts most recently
played, newest first. Scrolling the Recent tab loads older pages of Spotify's
play history, and each of those pages was noted at the front of that order. A
playlist last played yesterday moved above one played an hour ago, and a
playlist played both recently and long ago was pulled to the top by its older
play.

## What changed

`absorb_recents` handled every page with `note_recent_contexts`, which inserts
each context at index 0 so the newest play of the page ends up first. That is
right for the first page, the newest plays, but a page requested with a
`before` cursor holds plays older than everything the order already has.

`absorb_recents` now tells the two apart by whether the request carried a
cursor. The first page is noted exactly as before. An older page goes through
`note_older_contexts`, which appends the contexts the order does not have yet,
newest first, and leaves contexts already listed in the place a newer play gave
them. The sidebar filter (playlists, albums, Liked Songs) moved into
`is_sidebar_context` so both paths share it, and the existing limit of 60 is
now the named constant `RECENT_CONTEXTS_KEPT`, which the append path respects.

No documentation change: the guide already describes the order as recently
played, which is what it now keeps.

## Verification

New test `app::tests::older_history_pages_do_not_jump_ahead_in_the_sidebar_order`
goes through `Action::ReloadRecents`, `Action::LoadMoreRecents` and the
`ApiResponse::RecentlyPlayed` handler. The first page plays from `p1` then `p2`;
the older page plays from `p3`, an artist, and `p2` again. It then fills the
order to its limit and checks that another older page adds nothing. On `main`
it fails:

```
assertion `left == right` failed: older plays follow the newer ones
  left: ["spotify:playlist:p3", "spotify:playlist:p2", "spotify:playlist:p1"]
 right: ["spotify:playlist:p1", "spotify:playlist:p2", "spotify:playlist:p3"]
```

With the change it passes, as do the existing Recents paging tests and the
sidebar ordering tests in `ui::sidebar::ordering_tests`.

Run on Windows 11:

- `cargo fmt --all --check`
- `cargo clippy --locked --all-targets --no-default-features --features demo -- -D warnings`
- `cargo test --locked --no-default-features --features demo --lib` (619 passed)

The default MilkDrop feature needs GLEW, which this machine does not have, so
the default and all-features builds are left to CI. Not tested on macOS or
Linux; the change has no platform-specific code.


crmne:
Fixed on main in 8755546. Loading older history now keeps newer listens and current plays at the top of the Library. Tests pass on Linux, macOS and Windows. This is not released yet. Thanks for the fix!

PR #452, opened 2026-09-14
feat(ui): fade the player bar between song colours

luiscuellar31:
## Why

The player bar uses a 12% tint from the currently playing album art. Until now, that colour changed in a single frame whenever the song changed, which made the whole bottom bar visibly snap to a new shade.

Most other visual changes in the app already ease smoothly, including hovers, sidebar drop shifts and lyric highlights. The player bar was one of the remaining places where the colour still changed instantly.

There is no open issue for this. Before starting, I searched open and closed issues and pull requests for `accent`, `color transition`, `fade`, `gradient`, `animate color`, `player bar background` and `album art color`.

## What changed

`src/ui/player_bar.rs` now eases between tint colours before blending them into the player bar.

Instead of switching directly from one song colour to another, the three colour channels are animated with `ctx.animate_value_with_time`. This is the same egui helper already used for the sidebar drop shift, collection table row shift and lyric line highlight.

No new animation state is stored on `App`, since egui already keeps that state per animation id.

When there is no album-art tint, the animation targets `palette.panel`. This also means the player bar fades smoothly back to its normal colour when playback stops.

The final colour is still calculated with the same blend strength as before:

```rust
blend(panel, tint, TINT_STRENGTH)
```

`TINT_STRENGTH` now gives a name to the existing `0.12` value.

The fade duration is controlled by `TINT_FADE_SECONDS`, currently set to `0.45`. This is intentionally longer than animations such as the `0.12` second row shift because a full colour transition benefits from being slightly slower.

When there is no tint:

```rust
blend(panel, panel, TINT_STRENGTH)
```

still returns exactly `panel`, so the no-art case keeps the same final colour as before.

Another useful effect of using egui's animation system is that changing songs again while a fade is still running continues from the colour currently visible on screen. It does not jump back to the previous song's colour first.

The page header gradient, `page_tint`, is intentionally unchanged. That colour follows the page being viewed instead of the currently playing song, so it has separate behaviour and timing.

`README.md` was also updated in the existing Album-art colour section to document the new behaviour.

## Verification

I ran the following commands on macOS 15.6, aarch64 Apple silicon:

```sh
cargo fmt --all --check
cargo test --locked --all-targets
cargo test --locked --all-targets --all-features
node --test .github/scripts/issue-assessment.test.cjs
```

Everything passes.

The Rust test suite reports 670 tests with 0 failures, and the issue assessment suite reports 30 passing tests.

I did not build the change locally on Linux or Windows. It was compiled and tested on macOS only, although this change does not touch any platform-specific code.

New tests were added under `player_bar_tint_tests`. They use `egui::Context` with `RawInput::time`, which makes the animation deterministic instead of depending on the real system clock.

The tests cover three cases:

- `a_song_without_art_leaves_the_panel_untouched` checks that the normal panel colour stays unchanged in both dark and light themes.
- `the_first_frame_shows_the_tint_without_fading_in` checks startup behaviour so the player bar does not fade in from black on its first frame.
- `changing_songs_crosses_over_to_the_new_tint` checks that the first retargeted frame still shows the old colour, a later frame is between both song colours on every channel, and the animation eventually reaches the new colour.

### Before and after, recorded

All four videos were recorded in the real app using a Premium account. The baseline is `main` at `16ee665`, and the candidate is this branch.

For each pair, I skipped between the same songs in the same playlist, using the same theme, at 1920x1080 and 30 fps.

The important part to watch is the player bar at the bottom. Before this change, the colour switches in one frame after every skip. After this change, it fades smoothly from one song colour to the next.

**Dark theme, after this change**

https://github.com/user-attachments/assets/5dae7c07-8659-4ea1-90f1-4769671f1306

**Dark theme, before (`main`)**

https://github.com/user-attachments/assets/5b03760f-5ffc-4f80-8f6a-948e84bb9a3f

**Light theme, after this change**

https://github.com/user-attachments/assets/a6cda7d9-e4a6-40c3-af6f-3efb670e0ff8

**Light theme, before (`main`)**

https://github.com/user-attachments/assets/4a3e675b-bc83-4953-a3ef-df644546eb95

The recordings are intentionally silent because the audio does not provide any useful information about the colour transition, and removing it also keeps Spotify's audio stream out of the clips.


crmne (changes requested):
Visible change: a 450 ms player-bar tint fade. Please provide the required HTML comparison with the transition states.

Preserve the panel alpha when there is no artwork tint. The previous None branch returned palette.panel unchanged; the new unconditional blend uses an opaque color, so transparent custom themes change even with artwork tinting off. Add coverage with a translucent panel, plus a genuinely mid-fade retarget test.

luiscuellar31:
Thanks for the review. I’ve preserved the panel’s alpha when there’s no artwork tint, added tests for a translucent panel and a mid-fade retarget, and added the requested HTML comparison with the transition states. The comparison is linked in the PR description, and all CI checks now pass. Could you take another look and confirm the visual change?

docs/assets/player-bar-tint/after-dark-narrow-end.webp
crmne, line 1: Please don't commit screenshots meant for the PR.

crmne (changes requested):

luiscuellar31:
You're right, sorry about that. I misunderstood where the PR review screenshots should go. I've removed the folder from the branch history and moved the screenshots and standalone HTML comparison to PR attachments. The PR description now links the comparison. Thanks for pointing this out.

luiscuellar31:
I've pushed the sign-out fix in dc67c63. The player-bar tint animation now starts fresh after sign-out, and I added a regression test that uses the same egui context across sessions. All six CI checks pass.

Could you please take another look at the code and visual evidence?

crmne (approved):
Thanks! The alpha handling, the mid-fade retarget and the sign-out reset all look right, and the fade stops asking for frames once it settles. Merging.

PR #471, opened 2026-09-14
Ignore a playlist page asked for before the playlist list was reloaded

kevin9327:
## Why

Following or unfollowing a playlist, or saving a playlist's details, reloads the library's playlist list from the top (`load_playlists`). The only thing that stops a reload is a first page still on its way, so a later page asked for while the list pages through (for an account with more than 50 playlists) can still arrive after the reload. Nothing told it apart from the new list's pages:

- If it arrives before the new first page, `slot => *slot = Loadable::Loaded(page.items)` makes that later page the whole list, and its `next` asks for the page after it.
- The new first page then starts its own run of pages, and both runs extend the same list, so playlists repeat and appear out of order in the sidebar until the next reload.

## What changed

`Library` keeps `playlists_asked`, the offset of the later page on its way. Asking for a later page records it, taking an answer clears it, and a reload forgets it. An answer for a later page is only taken when it is the page on its way. First pages are unaffected: only a reload asks for one, and it refuses while one is outstanding.

## Verification

Windows 11, with `--no-default-features --features demo` because the default MilkDrop feature needs GLEW from vcpkg, which is not installed here:

- `cargo fmt --all -- --check`
- `cargo clippy --no-default-features --features demo --all-targets -- -D warnings`
- `cargo test --no-default-features --features demo --lib`: 757 passed, 1 ignored

The new test `a_playlist_page_asked_for_before_a_reload_is_not_taken` loads the first page, reloads through `ApiResponse::PlaylistFollowChanged` while the second page is on its way, delivers that late page, then the new pages and a repeated answer. On `main` the late page replaces the list being reloaded:

```
panicked at src\app.rs:9575:9:
the late page is not the reloaded list
```

With the fix but without clearing `playlists_asked` when an answer is taken, the repeated answer is appended again:

```
assertion `left == right` failed
  left: Some(["new", "a", "b", "c", "b", "c"])
 right: Some(["new", "a", "b", "c"])
```

It passes with the whole change. There is no visual change to capture.

- [x] I updated documentation for user-visible behaviour, settings, files, or network access. (No documented behaviour changes.)


crmne (changes requested):
There is no intended visual change. The reload guard is useful, but an offset does not identify which load a page belongs to. If the new first page requests offset 50 before an old offset-50 response arrives, that old response still matches playlists_asked and is accepted into the new list. Please carry a request generation through MyPlaylists and add that response-order regression, following the generation checks already used for other Library shelves.

PR #477, opened 2026-09-15
nix: improve packaging

vidhanio:
## Why

Bring nix package in sync with https://github.com/NixOS/nixpkgs/pull/563375, improving codesign flow and improving `projectm-sys` patch.

## What changed

Uses `substituteInPlace` to fix the projectm-sys patch issue rather than overriding `$CMAKE`. This is how patches to deps are usually done upstream in nixpkgs and is less brittle, as it will invalidate the substitution if something changes upstream that can allow us to drop this patch.

It also uses `rcodesign` to sign rather than relying on `/usr/bin/codesign` existing, another practice used in upstream nix packages.

Finally, I have also merged the `spotifast` and `spotifast-app` input into one derivation, as is the convention upstream.

cc @ankarhem, would appreciate if you could give the darwin logic in this and [the nixpkgs pr](https://github.com/NixOS/nixpkgs/pull/563375) to test to ensure it works as expected.

List the commands and platforms you actually tested. Include before/after
screenshots or a recording for UI changes.

```
nix build .#packages.x86_64-linux.fastpotify
nix build .#packages.aarch64-linux.fastpotify
```


vidhanio:
btw, we can just have the same package for `spotifast` and `spotifast-app`, just like how it is done upstream and how it is done in nixpkgs generally, let me know if it would be wanted to merge those (imo it is unexpected to have them be their own packages)

ankarhem:
darwin works as expected

I agree changing the packaging to be more in line (like the nixpkgs) would be an improvement.

PR #478, opened 2026-09-15
fix(macos): restore native titlebar double-click

### LINKED ISSUE #350: [MacOS] Reimplement native menubar behaviour
### Before requesting

By disabling the native menu bar not all actions which are assumed to be working works.

An additional gestrue should be added in addition to the drag. Double clicking should make window expand to take up the full available space _by expanding the window surface_, over ~0.2s. Double clicking it again should minimize the window to the previous size and location.

Considerations like os setting for reduced animation settings should be taken into account.

Afaict it will be necessary to simulate the animation by widening the window in steps

It is the expected native behaviour which should be reimplemented because the native bar was removed.

luiscuellar31:
Fixes #350.

## Why

Spotifast draws its own top bar on macOS, so double-clicking it did not trigger the native title-bar action configured in System Settings.

macOS users expect this area to behave like a native title bar and zoom, fill, minimise, or do nothing depending on their Desktop & Dock preference.

## What changed

- Treat the full top area, including the sidebar, as the macOS title bar.
- Read the system title-bar double-click preference and trigger the corresponding AppKit action.
- Defer the native action until egui finishes the current frame to prevent the interface from stretching during the window animation.
- Enable VSync on macOS so OpenGL buffer swaps stay synchronized with the native window animation and avoid flickering when restoring the window.
- Keep the implementation macOS-specific while preserving the existing Windows and Linux behaviour.
- Document the native macOS window behaviour in the README.

## Verification

Tested locally on macOS with Apple silicon:

- `cargo run --locked --no-default-features --features demo -- --demo`
  - Repeatedly double-clicked the top bar to zoom and restore the window.
  - Confirmed the native animation remains smooth without stretching or flickering at different window sizes.
  - Confirmed the interaction works above both the sidebar and the main content.
  - Confirmed the behaviour works correctly in both light and dark themes.
- `cargo check --locked --all-targets`
- `cargo fmt --all --check`
- `node --test .github/scripts/issue-assessment.test.cjs`

## Visual evidence

Both recordings include the light and dark themes.

**Before**

https://github.com/user-attachments/assets/0b0c9824-1410-4c32-b384-8963f04341bb

**After**

https://github.com/user-attachments/assets/6a41937f-18f1-440c-95a0-5fc060a742c2

The Jekyll documentation build was not run because Bundler 4.0.16 is not installed locally.


crmne (changes requested):
Visible interaction changes: the whole macOS top area becomes a draggable title bar, double-click follows system preferences, and macOS VSync is enabled.

Please rebase onto main, preserving the Touch Bar crash guard from #532. Supply the required HTML comparison and native checks for Fill, Zoom, Minimise and Do Nothing, plus hidden/minimized playback with VSync. The current mapping sends Maximize and Zoom to the same performZoom action, so the documented Fill behavior needs to be demonstrated or implemented separately.

luiscuellar31:
I’ve updated the PR description with the HTML comparison and results for Fill, Zoom, Minimise, Do Nothing, and playback while hidden or minimized with VSync. The screenshots show the unchanged resting UI; the recordings make the interaction change clearer. Native checks and the focused local test passed. The branch preserves the Touch Bar crash guard from #532.

PR #482, opened 2026-09-15
fix(macos): prevent player window switch crash

### LINKED ISSUE #450: Crash when switch between Big and Small player
### Spotifast version and how you installed it

fastpotify-v0.8.0-macos-universal.dmg

### Operating system

macOS

### System details

macOS 15.7.9

### Spotify plan

Premium

Switching from the main window to the Winamp mini player crashes the app instantly.

Steps to reproduce:
1. Start Fastpotify, main window opens.
2. Press Cmd+Shift+M (the shrink button next to the settings gear does the same).
3. The app dies before the mini player is drawn.

Expected: the window switches to the mini player.
Actual: the process is killed with SIGILL. macOS shows the standard crash dialog.

Reproducible: every time

The macOS crash report is attached.
[apple-report.txt](https://github.com/user-attachments/files/32198209/apple-report.txt)

I can test patches on this hardware if that helps.

### fastpotify.log

```text
[2026-09-14T14:40:57Z WARN  fastpotify::api::client] Spotify rate limit source=shared wait=12s
[2026-09-14T14:40:57Z INFO  fastpotify::api::client] Spotify cooldown source=shared duration_ms=12000
```

luiscuellar31:
Fixes #450.

## Why

Switching between the main window and the mini player could crash Spotifast on macOS. AppKit attempted to clean up a Touch Bar observer after the previous native window had already been destroyed.

## What changed

- Hide macOS windows before `eframe` destroys their native views.
- Apply the workaround from the common window shutdown path, covering both directions of the main and mini player switch.
- Keep the change macOS-specific. Windows and Linux retain their existing behavior.
- Enable the existing AppKit `NSWindow` API needed by the workaround.

## Verification

Tested on macOS with Apple silicon:

- `cargo run --locked --no-default-features --features demo -- --demo`
  - Switched between the main window and mini player eight consecutive times.
  - Confirmed both window modes continued opening correctly.
  - Confirmed the application closed normally afterward.
- `cargo check --locked --all-targets`
- `cargo fmt --all --check`
- `node --test .github/scripts/issue-assessment.test.cjs`

The affected Intel Mac with Touch Bar was not available, so the fix was smoke-tested on Apple silicon only.

The packaging and Jekyll checks could not complete locally because the required Linux packaging tools and Bundler 4.0.16 were unavailable.

- [x] Documentation is not required because this fixes a crash without changing user-visible behavior, settings, files, or network access.


crmne:
The Touch Bar cleanup crash is now addressed on main by ea5bd789e9f099490fa402a6ee8ec42ef3859797 (#532), which targets the exception from #450. Closing this alternative workaround as superseded. Thanks for investigating it. The fix is not released yet.

PR #511, opened 2026-09-17
Keep a rootlist request that arrives before the engine

### LINKED ISSUE #291: Add an option to add songs to a collaborative playlists
### Before requesting

When i try to add a new song to a collaborative playlist, the app doesn`t have the option to do that, only can add to playlists of the user account.

Like the original Spotify app. 
With right click > add playlist > select 

_Spotify app_
Image

 
_Fastpotify app_
Image

Give users the possibility to add songs to a collaborative playlist.

Magniquick:
Fixes #291.

A playlist shared by invitation is editable only by the rootlist's word, and the rootlist request is sent once, from `handle_api` when `MyPlaylists` finishes paginating. `fetch_rootlist` drops that request when the engine has not connected yet:

```rust
fn fetch_rootlist(&self) {
    let Some(engine) = self.engine.clone() else {
        return;
    };
```

A cached web token regularly finishes the playlist library about a second before librespot authenticates, so the request lands in that gap. Nothing sends it again, `editable_by_grant` stays empty for the session, and `can_edit_playlist` falls back to `owned || collaborative`. Both are false for an invitation share, so the playlist never reaches Add to playlist.

This holds the request and runs it where `self.engine` is assigned.

On my account the ordering is reproducible from a cold start, and the fix recovers it. Instrumented with temporary log lines:

```
rootlist deferred, engine not ready
engine ready, retrying deferred rootlist
rootlist fetch starting
rootlist arrived, 33 entries, 4 editable by grant
```

Four matches what `examples/rootlist_probe` reports for the account, one shared by invitation and three owned. The playlist then takes songs from the menu.

`cargo test --no-default-features` passes, 741 unit tests plus the integration binaries. `cargo fmt --check` and `cargo clippy` are clean.

`CachedRootlist` persists `entries` but not `editable`, so a start that wins the race leaves nothing behind for the next one. `a_playlist_shared_by_invitation_takes_songs` feeds `Event::Rootlist` straight to the app and skips the backend, which is why the suite stays green.


crmne (changes requested):
Incorporated into main as 91fa86e, with contributor credit and regression coverage for replaying the waiting read once and discarding work on sign-out. The connection reference is updated. Thanks for finding and fixing the startup race.

PR #512, opened 2026-09-17
linux: Add `spotifast like` support

mgsloan:
## Why

The MPRIS protocol does not have a "like" action, and so spotifast cannot be remote controlled to like the current track.

## What changed

This adds support for liking tracks via the Spotifast-specific D-Bus interface.  `spotifast like` cli now sends a message to that interface.

Initial change created with AI.  I manually simplified and polished.

## Verification

Debian.  Verified that running `spotifast like` works.

I did not include tests because it doesn't seem worth the weight.  I don't think docs need an update.


crmne (changes requested):
The Linux like command fits and keeps the existing private D-Bus interface. Please add a private-bus regression test proving ToggleSaved reaches the running app without raising or starting a window, and cover the no-running-instance error. Update the CLI documentation to say that like is available on Linux and toggles the current song’s saved state. Those behavior and documentation checks are required for this new command path.

mgsloan:
Makes sense!  I've made some changes to address the feedback

PR #525, opened 2026-09-17
fix(ui): stop shuffle button from starting playback

luiscuellar31:
## Why

The Shuffle button next to **Play** used to start playback immediately when clicked.

That made it impossible to enable shuffle first and decide separately when to start playing the playlist. Since the button is meant to control shuffle mode, it should only toggle that state and leave playback to the **Play** button.

## What changed

- Updated the Shuffle button next to **Play** so it only toggles shuffle mode and no longer starts playback.
- Kept the active shuffle state visible with the accent colour.
- Left the explicit **Shuffle play** action in context menus unchanged.
- Added a regression test to make sure toggling shuffle does not emit a playback action.
- Updated the README and user guide to reflect the new behaviour.

There is no linked issue.

## Verification

Tested on macOS with:

```sh
cargo fmt --all --check
cargo test --locked ui::collection::tests
cargo test --locked --all-targets
cargo test --locked --all-targets --all-features
```

All of the commands above passed.

I did not include screenshots because this change does not affect the interface's appearance or layout. The behaviour change is covered by the regression test.

I was not able to run the local documentation site build because Bundler 4.0.16 is not installed in this environment. The packaging launcher tests also depend on Linux-compatible tooling and a newer Ruby version than the one available on this macOS setup.

- [x] I read `CONTRIBUTING.md` and kept this pull request focused on one concern.


crmne (changes requested):
Visible interaction change: the collection Shuffle button becomes a global mode toggle and no longer starts that collection.

Please add integration coverage for toggling Shuffle before any playback device is active, then pressing Play, and for toggling it while a different collection is playing. The current test only checks the emitted Action; it does not establish that the selected mode survives until Play. Include the required HTML comparison and update the queue reference alongside the user guide. Visual-scope approval is still pending.

luiscuellar31:
Thanks for the feedback! I’ve added the tests and updated the queue reference. The HTML comparison and screenshots are now in the PR description too.

crmne (changes requested):
The look is approved: Shuffle beside Play is now a toggle, the same as Spotify's desktop app. Thanks for adding the tests and the queue reference.

Two things before this can merge:

1. With no device active, turning Shuffle on from the playlist page lights the page's button, but the player bar still shows Shuffle off, because it reads shuffle only from the playing song (`src/ui/player_bar.rs` around line 273). Its toggle in `src/app.rs` (around lines 7496-7498) reads that same "off" value, so clicking it turns shuffle on again and nothing visibly changes. Media keys and the command line behave the same way. When nothing is playing, both the player bar and its toggle should use the saved Shuffle choice. Please add a test that turns Shuffle on from the page with no device, then checks the player bar state and one player-bar toggle.
2. In the README, the new Shuffle text was inserted in the middle of the playlist bullet (around lines 134-138), so the Refresh and checkpoint lines that follow now read as part of it. Please move it after the end of that bullet.

luiscuellar31:
Thanks, both fixes are in bd3b85a. The player bar now uses the saved Shuffle choice when nothing is playing, with a test covering the playlist and player bar toggle. I also moved the README text after the full playlist bullet.

crmne:
Thanks @luiscuellar31, both fixes look right. Squashed onto main as 29efec5 with you as the author. It will ship in 0.10.

PR #528, opened 2026-09-18
Add interactive MacBook notch Now Playing overlay widget for macOS

saikumargudelly:
### Why this belongs in Spotifast

On MacBooks with a camera notch, the bezel area around the notch is unused. This adds a native macOS overlay widget that sits directly under the notch when Spotifast is playing music in the background or minimized. Hovering near the notch reveals a compact "Now Playing" floating island with album art, dynamic audio visualizer, a wavy progress bar matching the track's color, and quick playback controls.

### User-visible changes

- **Notch overlay**: A floating capsule drops down under the physical notch when hovering over the notch area while the app is in the background. It smoothly retracts when the cursor leaves.
- **Frosted glass styling**: Styled to match iOS Control Centre cards using native dark blur vibrancy, a clean translucent tint, borderless contours, and a subtle top specular sheen.
- **Dynamic audio visualizer**: A 5-bar equalizer indicator that dynamically dances to real-time audio frequencies (bass to treble) extracted from the player's `AudioTap` and `Analyser`. The bars rest flat at baseline height when music is paused or stopped.
- **Wavy progress bar**: Displays an animated wave while playing (straight capsule when paused) that matches the dominant color of the current album art. Supports click-to-seek.
- **Track transition**: Smooth crossfade animation when skipping or switching tracks instead of abrupt cuts.
- **Playback controls**: Star (save to library), previous, play/pause, next, and bring-to-front.
- **Settings toggle**: Added "MacBook notch widget" switch in Settings under "Playback on this computer" (macOS only, default enabled).

### Implementation details

- Uses AppKit `NSVisualEffectView` with `HUDWindow` material and `BehindWindow` blending for native frosted glass vibrancy.
- Anchored at `y = notch_h + 8.0` so it rests completely below the physical camera housing without overlapping.
- Feeds 512 mono audio samples directly from the player's `AudioTap` into `Analyser::step` to compute real-time spectral energy for 5 frequency bands (bass, low-mids, midrange, upper-mids, treble) with smooth falloff decay.
- All macOS-specific code is isolated in `src/mac_notch.rs` behind `#[cfg(target_os = "macos")]`.
- Cross-platform facade in `src/notch.rs` provides zero-overhead no-op stubs on Linux and Windows.
- Repaint rate is throttled to ~30 FPS only while the widget is actively expanded or animating; idle state uses 0 extra cycles.
- No new crates added. Reuses existing `objc2` and `objc2-app-kit` dependencies.

### Checks performed

- `cargo fmt --all --check`
- `cargo test --locked --all-targets` (all 779 tests pass)
- Tested locally on macOS at normal and narrow display sizes.


saikumargudelly:
Addressed the review feedback in commits 23693e0 and 0151e2b:

- **Analyzer and audio tap gating**: FFT sampling and window extraction are now gated behind `enabled && active`, returning flat baseline bars during idle without allocations. Remote playback visualizer levels stay flat when no local audio tap is present.
- **Physical notch detection**: Displays without a physical notch (`top_inset <= 0`) return `None` and do not create an overlay window.
- **Hover expansion when paused or stopped**: The collapsed overlay remains registered for cursor hover while Spotifast is in the background, allowing users to hover over the notch to see the player and resume playback even when music is not currently playing.
- **VoiceOver accessibility**: Added explicit accessibility descriptions to all buttons and SF Symbols.
- **Podcast episode handling**: Hid the save button and suppressed `ToggleSaved` for episodes.
- **Async artwork loading**: Offloaded artwork file reads to a background thread and reset the cached path when playback clears.
- **Tray mode refresh rate**: Wakes the headless tray loop at 33ms (~30 FPS) while the notch widget is active.
- **Tests and documentation**: Added settings serialization tests, pure notch geometry and state tests, and updated the user guide and settings reference.

saikumargudelly:
All review feedback addressed, formatting and glass appearance polished, and all 6/6 CI checks have passed. Ready for review.

saikumargudelly:
Addressed all 5 findings from the latest Copilot review (commit 53aef12):

- **Remote visualizer**: bars now stay flat when `now.local == false`. The sine-wave fallback fabricated audio activity that the widget never measured.
- **Idle repaint timer**: `request_repaint_after` is now gated on `active` (expanded or animating). The prior unconditional 200 ms tick added ~5 extra frames/second for every macOS user even when the card was idle. Hover and state-change paths already call `notch::wake()` so the idle tick was redundant.
- **Overlay without a track**: `orderFrontRegardless` is now guarded by `ctrl.track.is_some()`. The invisible window no longer intercepts clicks at the notch when nothing is playing.
- **Artwork race**: replaced the single `PENDING_ART` slot with an `ART_GENERATION` `AtomicU64` counter. Each spawned read-thread captures the generation at launch and only writes `PENDING_ART` if no newer load has started since. Fast track changes no longer lose the latest artwork.
- **Progress bar accessibility**: the canvas view now carries `AXSlider` role and `Playback Position` label so VoiceOver announces and can navigate the seek control.

saikumargudelly:
Screenshot 2026-09-19 at 10 16 19 AM

Visual Evidence (macOS Notch Widget)
Here is a capture of the expanded notch card in active playback with:
- Album art and track metadata
- Honest visualizer bars (FFT-driven from local AudioTap, flat when no stream data)
- Accessible seek bar (`AXSlider` role) and playback controls

saikumargudelly:
Addressed review feedback and final edge-case hardenings:

1. **Minimized Window Dispatch Timers:** Switched hover expand/collapse delays to native Grand Central Dispatch timers (`dispatch_after_f`) running directly on `_dispatch_main_q`. Minimized/throttled eframe main windows no longer stall notch interaction.
2. **FFI Unwind Safety:** Wrapped GCD callbacks (`on_artwork_loaded`, `on_hover_timer_fired`) with `std::panic::catch_unwind` to prevent fatal Rust panics from crossing the C FFI boundary.
3. **Accessibility & VoiceOver:** Configured AppKit `setAccessibilityElement: true` on `FastpotifyCanvasView` under the `AXSlider` role to support screen readers safely without iOS-specific selector panics.
4. **Optimistic Star & Play Controls:** Immediately toggles state and updates UI icons on click before dispatching backend actions, adhering to the optimistic UI contract without flicker.
5. **Memory & Artwork Protection:** Added a 10MB file metadata guard in `spawn_art_loader` and generation token checks to drop stale or oversized artwork files.
6. **Zero Collapsed Overhead:** Stripped out layer redraws and animation increments when collapsed, guaranteeing zero GPU draw cycles.
7. **Safe Dynamic Class Lookups:** Switched static `objc2::class!` macros for `CAMediaTimingFunction` and `NSImageSymbolConfiguration` to panic-free `AnyClass::get` runtime lookups.
8. **Precision Loss & Contrast:** Clamped animation phase via modulo to avoid floating-point precision loss over time, and added unit tests validating luminance contrast boosting.

saikumargudelly:
Addressed review feedback, comprehensive micro edge-case audit, and code deduplication (commit 4163e15):

1. Rebased on latest main: Kept linear history on top of commit 6ce15a9 with zero merge commits.
2. Touch Bar crash guard: Restored mac_touchbar_crash_guard in src/entrypoint.rs to maintain protection against AppKit Touch Bar KVO observer exceptions on window close.
3. macOS 11 runtime safety: Guarded NSScreen::safeAreaInsets with respondsToSelector so macOS 11 returns zero insets and never sends an unrecognized selector.
4. README documentation: Restored artist profile navigation documentation while retaining the notch widget section.
5. State machine & regression tests: Extracted window visibility, hover timer transitions, controller active checks, and incremental changes into pure helpers in src/notch.rs. All 794 unit tests pass across platforms.
6. Micro edge-case hardenings:
   - Artwork load transition elimination: Excluded art_path from is_track_metadata_different. Delayed artwork file loading now flows strictly through incremental state changes without erroneously triggering a 220ms track transition crossfade on the active song.
   - Unconditional collapse reset on hide: Ensured that ordering the window out while a collapse animation is mid-flight unconditionally resets view alpha, card frame, and window frame to physical notch dimensions, preventing the window from re-appearing at expanded dimensions on subsequent playback.
   - Display inset single traversal: Streamlined screen top inset and geometry discovery in compute_frames into a single find_map pass.
   - Tabular digit font for time labels: Switched elapsed and remaining time labels to monospacedDigitSystemFontOfSize:weight: with system font fallback, eliminating horizontal digit width jitter every second.
   - Dynamic VoiceOver slider value & artwork accessibility: Dynamically set accessibilityValue on the progress canvas (e.g. "1:23 of 3:45") and configured explicit accessibilityLabel ("Album artwork") on art_view.
   - Canvas placeholder invalidation: Added ctrl.canvas_view.setNeedsDisplay(true) on background artwork completion so the placeholder disc cleanly transitions to the loaded album art.
   - Incremental duration & accent updates: Tracked duration boundary crossings and tint changes in detect_incremental_changes to refresh remaining times and accent tint when stream metadata arrives.
   - String sanitization & title fallback: Stripped newline and carriage return characters from title and artist fields to protect single-line NSTextField rendering, and added a Spotifast fallback for empty track titles.
   - Root view frame origin: Initialized the root FastpotifyNotchView frame with local origin NSPoint::ZERO rather than screen coordinates before setting as window content view.
7. Line-by-line deduplication and code cleanup:
   - Removed dead helper should_expand_notch from src/notch.rs (superseded by should_schedule_expand and should_perform_expand).
   - Consolidated should_show_save_button into can_toggle_saved in src/notch.rs, eliminating the redundant duplicate function.
   - Extracted shared button setup and action wiring helpers (setup_icon_button, attach_button_action, set_button_tint) in src/mac_notch.rs, deduplicating repetitive AppKit button configuration in init() and re-using update_play_button_ui and update_star_button_ui.
   - Replaced redundant 8-field assignment in sync_state incremental updates with whole-struct clone.
   - Extracted apply_pending_artwork_locked, set_view_corner_radius, configure_single_line_label, reset_track_fade, and update_artwork_path to eliminate repeated blocks.
   - Simplified idle_duration in src/entrypoint.rs by taking advantage of cross-platform notch::is_active().
   - Streamlined FFT analyzer level sampling and band aggregation in src/app.rs into a concise band iteration loop.
   - Eliminated redundant reset_time_labels by reusing update_time_labels with zero parameters, and streamlined waveform bar heights in handle_draw_canvas.

crmne (changes requested):
Visible scope: a new macOS notch overlay, enabled by default, with its own controls, animation, visualizer and Settings toggle. This scope has not received maintainer visual approval.

Please provide the required HTML comparison with matching light/dark and narrow/normal captures, including collapsed, expanded, paused and remote-playback states. Also correct the zero-collapsed-overhead claim: sync_notch_widget still computes the FFT whenever enabled and playing locally, even while collapsed or in the foreground. Gate that work on the visible visualizer and cover the transition in tests.

PR #544, opened 2026-09-19
zh-Hans: align repeat/queue/lyrics wording with the UI they appear in

huojian17-star:
Ten `msgstr` in `assets/i18n/zh-Hans.po`. No `msgid`, placeholder, or structural change; `msgfmt --check --check-format` passes.

I read the call sites before changing anything, and several of these are not `theme::text` labels but `theme::icon_button` tooltips. That function's last argument is both the screen-reader name and the hover text:

```rust
response.widget_info(|| egui::WidgetInfo::labeled(WidgetType::Button, ui.is_enabled(), tooltip));
response.on_hover_text(tooltip);
```

So the wording has to work without any visual context, and it is not width-constrained.

### Repeat group

`player_bar.rs` states the mode a click *enters*, not the current one:

```rust
let (repeat_icon, repeat_color, tooltip) = match repeat {
    RepeatMode::Off     => (Icon::Repeat,  dim,            gettext(.., "Repeat")),
    RepeatMode::Context => (Icon::Repeat,  palette.accent, gettext(.., "Repeat one")),
    RepeatMode::Track   => (Icon::Repeat1, palette.accent, gettext(.., "Repeat off")),
};
```

`RepeatMode` maps straight onto the Web API's `off` / `context` / `track`. The three tooltips are one set, so they should read as one set:

| msgid | was | now |
|---|---|---|
| `Repeat` | 重复播放 | 列表循环 |
| `Repeat one` | 单曲循环 | 单曲循环 (unchanged) |
| `Repeat off` | 关闭重复播放 | 关闭循环 |

### Queue

| msgid | was | now |
|---|---|---|
| `Queue` | 播放队列 | 播放列表 |
| `Clear queue` | 清空队列 | 清空播放列表 |
| `Queue {date}` | 队列 {date} | 播放列表 {date} |

### Empty states

Both are `empty_state` headings, so they sit above a smaller subtitle and should match each other:

| msgid | was | now |
|---|---|---|
| `Nothing playing` | 未在播放 | 暂无播放曲目 |
| `Nothing queued` | 队列为空 | 暂无待播歌曲 |

### Lyrics

| msgid | was | now |
|---|---|---|
| `Follow` | 跟随 | 跟随歌词 |
| `Enable lyrics motion` | 启用歌词动画 | 启用歌词动效 |
| `Reduce lyrics motion` | 减少歌词动画 | 减弱歌词动效 |

`Enable` and `Reduce` are the two states of one toggle (`lyrics.rs:351`), so both halves are changed together. `Reduce` is a reduction, not an on/off: with it set, the scroll animation duration goes to `0.0` and the lyrics still scroll (`lyrics.rs:497`).

### Not changed

- `Recent` (queue tab) stays 最近播放: it sits beside `Queue` and the two are now both four characters.
- `Playing next` / `Next up` stay 接下来播放 / 后续曲目: these are section headings with a list of songs under each, not buttons.

### One note on the checks

`.github/scripts/update-translations.sh --check` compares the committed `fastpotify.pot` against a fresh `xgettext` run. On current `main` those two already differ by source line numbers only: `app.rs` has moved about 161 lines since the template was generated, `sidebar.rs` two. So that step fails before this PR is involved. I left the template alone to keep the diff to the ten strings; happy to regenerate it here if you would rather the check be green.


huojian17-star:
A note on where the wording comes from, since the diff alone does not say.

The Chinese catalogue was machine-generated in one batch (all twelve locales share `PO-Revision-Date: 2026-09-09 13:30+0000`), so it reads as a literal rendering of the English rather than as a set of names chosen for the controls they sit on. The changes above are mostly about making the three groups read as groups.

**Repeat group.** The three tooltips describe the mode a click enters, so they form one cycle and should be named as one set. All three now end in 循环 (loop): 列表循环 / 单曲循环 / 关闭循环. The previous wording mixed a verb phrase (重复播放) with a noun phrase (单曲循环) and a negation (关闭重复播放), which reads as three unrelated things rather than one control cycling through three states.

I considered 顺序播放 for `Repeat` and did not use it. 顺序播放 is a real term in Chinese players, but it does not contain 循环, so it would break the set; and it is used inconsistently across apps to mean either "play through and repeat" or "play through and stop", which is exactly the distinction `off` and `context` draw. Keeping 循环 in all three names makes the axis explicit.

**Queue.** 队列 is a data-structure word. 播放列表 is what the rest of the catalogue already uses for a list of songs in order, and it also matches 清空播放列表 and 播放列表 {date} now that all three changed together. `Playlist` is separately 歌单, which is the standard Chinese term for a saved playlist, so the two do not collide.

**Empty states.** `Nothing playing` and `Nothing queued` are both `empty_state` headings — large text above a smaller subtitle. 未在播放 was a literal rendering that reads as a status code; 暂无播放曲目 matches the register of the subtitle beneath it. 队列为空 was changed to 暂无待播歌曲 so the pair is parallel and so the queue wording follows the rest.

**Lyrics.** 跟随 alone does not say what is being followed; the button appears after a manual scroll and resumes auto-scrolling, so 跟随歌词 names the object. For the motion toggle, `Reduce` is a reduction rather than an off switch — with it on, the scroll animation duration becomes `0.0` and the lyrics still scroll — so 减弱 is more accurate than 减少, and 动效 covers both.

These are judgement calls about Chinese, not corrections of factual errors, and I am happy to change any of them if you would rather use different terms. The one I would flag as most open to debate is 列表循环 versus 顺序播放.

huojian17-star:
Thanks — two of the three are fixed in dbd816a.

**Queue terminology.** You're right, and I've reverted it. `播放列表` was my change and it was wrong for the reason you give: `Queue` is the active playback queue, while `Save as a playlist` is a separate action, so naming the former 播放列表 makes those controls claim to open a saved playlist. `Queue`, `Clear queue` and `Queue {date}` are back to 播放队列 / 清空队列 / 队列 {date}. `Playlist` stays 歌单.

**Stale template.** Also fair. I had described the drift in the PR body and left it alone to keep the diff to the ten strings, but that reasoning doesn't hold: the check is a requirement on the PR regardless of who made it red, and refreshing the template is a small change. `fastpotify.pot` is regenerated in the same commit — six `#:` reference lines in `app.rs` had drifted by about 161 lines since the template was generated on 2026-09-15.

Verified locally, following what the script does:

- the committed template now matches a fresh `xgettext` run over `assets/i18n/POTFILES`, ignoring only the `POT-Creation-Date` line
- `msgfmt --check --check-format` passes on all thirteen catalogues

**UI review evidence.** This one I can't satisfy yet, and I don't want to fake it. The policy asks for same-state before/after captures in demo mode across light/dark and narrow/normal, which means building the app and running `--demo --demo-language zh-Hans`. My machine has no MSVC linker (rustup installed, `link.exe` missing), so a build fails at the link step. I'd rather say so than attach screenshots I didn't take.

If you'd like the evidence before merging, I can install the build tools and produce it — just say so and I'll do it. Otherwise, the two changes above are the ones I could verify without a compiler, and the screen-reader names you flagged (`Mute`/`Unmute`, `Close`, the two `(%)` sliders, `Hide sidebar`) are the ones I'd want to check visually first, since they're the ones with no visual context for the user.

huojian17-star:
Visual evidence, as the UI-review policy asks for. Captures were taken with the app's own demo mode, so the data, page and window size are identical within each pair — only the language differs.

`en` is the before (the catalogue falls back to English), `zh-Hans` is the after.

**Queue panel** — `Queue`, `Clear queue`, `Nothing queued`:

| Before | After |
| --- | --- |
|  |  |

**Lyrics panel** — `Follow`, and the `Enable`/`Reduce lyrics motion` toggle in the header:

| Before | After |
| --- | --- |
|  |  |

[Light/dark and window-size comparison — HTML with selectors, plus all 32 captures](https://github.com/huojian17-star/spotifast/archive/d30d7aa214eeaab158f7abbc0f7ce1891510c433.zip)

### How they were produced

```sh
spotifast.exe --demo --demo-language <en|zh-Hans> \
    --demo-show <fixture>[,light|dark] \
    --demo-size <760x620|1100x800> \
    --demo-shot <path.png>
```

`--demo-shot` writes the window's own framebuffer to a PNG and exits, so nothing outside the app window is captured. Built with `cargo build --release --no-default-features --features demo`.

Two things worth flagging about the build:

- **Release matters here.** egui's `warn_on_id_clash` is `cfg!(debug_assertions)`, so a debug build paints a red `First use of widget ID` box over the demo window. It has nothing to do with the translations — the English captures show it identically — but it makes the screenshots useless, so these are release builds.
- `--no-default-features` drops MilkDrop, which needs CMake, LLVM and vcpkg on Windows. None of the surfaces in this PR involve MilkDrop.

**Platform:** Windows 11 x86_64, NVIDIA RTX 4060, `--demo-shot`. Captures are from Windows only; I have no macOS or Linux machine to compare against.

### What the captures do and do not show

Visible: `Queue`, `Clear queue`, `Queue {date}`, `Nothing queued`, `Nothing playing`, `Follow`, and the `Repeat` group on the player bar.

Not visible, because they are icon-button tooltips that only appear on hover: `Repeat` / `Repeat one` / `Repeat off`, `Mute` / `Unmute`, `Close`, `Hide sidebar`, and the two `(%)` slider names. `Enable`/`Reduce lyrics motion` is one of these — the button is visible in the lyrics header, but its label is not. If you want those captured too, I can take hover shots; it just needs a different mechanism than `--demo-shot`, which exits before any hover can happen.

crmne (changes requested):
Visible change: seven Simplified Chinese labels and tooltips. The queue terminology correction looks good, and CI passes.

The new screenshots compare English with Chinese on the candidate build. Please compare the base and this PR with zh-Hans selected in both, at matching light/dark and narrow/normal sizes. Include the changed repeat and lyrics tooltips and empty states in the HTML comparison. That will show the wording this PR actually changes.

huojian17-star:
Point taken on all three. Here's where each stands.

**Queue scope.** You're right, and this was my error in the description rather than in the file. An earlier revision changed `Queue` / `Clear queue` / `Queue {date}` to 播放列表 / 清空播放列表 / 播放列表 {date}; review pointed out that `Queue` is the active playback queue while `Save as a playlist` is a separate action, and that was correct, so `dbd816a` reverted all three. They are byte-identical to base now and are not part of the diff. I've rewritten the PR description to say so and to list the seven strings that actually change.

**Visual evidence.** Replaced. The previous set compared English against Chinese, which showed that a translation exists rather than what this PR changes. The new set is **base vs this PR, both `--demo-language zh-Hans`**, at matching light/dark and narrow/normal sizes, and it includes the repeat and lyrics tooltips and the empty states.

[base vs this PR — HTML comparison with light/dark and narrow/normal selectors, plus all 36 captures](https://github.com/huojian17-star/spotifast/archive/c8472c6ef00ad194bcf8b26985dc553d7c6912a5.zip)

**Repeat tooltip** — 重复播放 → 列表循环:

| Base — 6ce15a9 | This PR — dbd816a |
| --- | --- |
|  |  |

**Nothing playing** — 未在播放 → 暂无播放曲目:

| Base | This PR |
| --- | --- |
|  |  |

Both sides were built with `cargo build --release --no-default-features --features demo`. Release matters here: egui's `warn_on_id_clash` is `cfg!(debug_assertions)`, so a debug build paints a red "First use of widget ID" box over the demo window — it appears identically in English and Chinese, so it is not related to the catalogue, but it makes the captures unusable.

**One thing I could not do with `--demo-shot` alone.** `Repeat`, `Repeat one`, `Repeat off`, `Enable lyrics motion` and `Reduce lyrics motion` are `theme::icon_button` tooltips; they only render while the pointer is over the button, and `--demo-shot` writes the framebuffer and exits before any hover can happen. I captured those five with a **local, uncommitted** hook that calls `Response::show_tooltip_text` — egui documents it as "always show this tooltip, even if disabled and the user isn't hovering it" — for one named string at a time. The hook is not in the PR; the diff is still only `assets/i18n/zh-Hans.po` and `assets/i18n/fastpotify.pot`. If you'd rather the evidence came from an unmodified build, the honest alternative is that those five tooltips can't be captured with the current `--demo-shot` path at all, and I'd rather say that than imply otherwise.

**Template.** Regenerated in `dbd816a`. The committed `fastpotify.pot` now matches a fresh `xgettext` run over `assets/i18n/POTFILES`, ignoring only the `POT-Creation-Date` line, and `msgfmt --check --check-format` passes on all thirteen catalogues.

huojian17-star:
Rebased onto current `main` and resolved the conflict.

**What the conflict was.** `main` moved 30 commits ahead of where this branch started, and two of them touched the same files: `a4a1bc61 Remove the lyrics motion button` and `06bc3cf2 Complete Spotifast rename`. The diff is now exactly one file — `assets/i18n/zh-Hans.po`, 5 insertions, 5 deletions — against `main` as of `5d80b8a`.

**The branch got smaller, and it's your change that did it.** `a4a1bc61` removed the lyrics motion button and left `Enable lyrics motion` / `Reduce lyrics motion` obsolete (`#~`) in every catalogue. There is no control left to translate, so those two are out of this PR. What remains is five strings:

| msgid | main | this PR |
|---|---|---|
| `Nothing playing` | 未在播放 | 暂无播放曲目 |
| `Repeat` | 重复播放 | 列表循环 |
| `Repeat off` | 关闭重复播放 | 关闭循环 |
| `Nothing queued` | 队列为空 | 暂无待播歌曲 |
| `Follow` | 跟随 | 跟随歌词 |

**Template.** No change needed — `a4a1bc61` regenerated `spotifast.pot` and the committed template is current on `main`, so this PR no longer touches it.

**Visual evidence.** Redone against the current `main` rather than the stale merge base, still with `zh-Hans` on both sides:

[main vs this PR — HTML comparison with light/dark and narrow/normal selectors, plus all 32 captures](https://github.com/huojian17-star/spotifast/archive/f201d034bfeadd86a4b9f0839386383753cb73bd.zip)

**Repeat tooltip** — 重复播放 → 列表循环:

| main — 5d80b8a | This PR — 41ccd47 |
| --- | --- |
|  |  |

**Follow** — 跟随 → 跟随歌词:

| main | This PR |
| --- | --- |
|  |  |

You'll notice the lyrics panel has no motion toggle on either side now — that's `a4a1bc61`, not this branch.

**One caveat I want to keep on the record.** The two repeat tooltips only render while the pointer is over the button, and `--demo-shot` writes the framebuffer and exits before any hover can happen. I captured them with a local, uncommitted hook that calls `Response::show_tooltip_text` for one named string at a time; the hook is not in the PR. If you'd rather the evidence came from an unmodified build, the honest alternative is that those tooltips can't be captured through the current `--demo-shot` path at all — I'd rather say that than imply otherwise.

Thanks for the earlier review notes — the Queue one in particular was right, and reverting it is why that group is no longer in the diff.

PR #551, opened 2026-09-21
Fix softened cover reload loop

hyperpuncher:
## Why

#517 introduced cover reload loop with increased idle cpu usage

## What changed

blurred covers now use the existing cached image instead of reloading it

## Verification

tested on arch

idle cpu usage dropped from 2% back to 0%


crmne:
Fixed on main in d1f093a. Softened covers now reuse the decoded artwork, including when the disk cache cannot be written. Thanks for finding the reload loop. Not released yet.

